PR #179, opened 2026-04-22
fix(manager_loop): polish bundle — issues #171, #172, #174, #175, #176

### LINKED ISSUE #171: bug(adapter): reject ':' in steer_context keys for DOT round-trip safety
## Problem

Surfaced during PR #170 review-squad pass. Dippin-lang's spec says \`steer_context\` map keys may not contain \`:\` because the block-form \`.dip\` formatter emits \`key: value\` lines and a key with \`:\` would break the round-trip. Tracker's \`flattenSteerContext\` (encoder) and the handler's \`parseSteerContext\` (decoder) don't reject or warn on such keys — they percent-encode only \`,\`, \`=\`, and \`%\`.

A pipeline author hand-authoring a DOT file with \`steer_context: "my:key=val"\` would get silent data corruption on any round-trip through block-form \`.dip\`.

## Proposed change

Add validation in either:
- \`pipeline/dippin_adapter.go::flattenSteerContext\` — fail loudly at adapter time
- \`pipeline/handlers/manager_loop.go::parseSteerContext\` — drop/warn on keys containing \`:\`

Either location works; the adapter-side rejection is probably better because it fails at graph-build rather than at runtime.

## Acceptance criteria

- [ ] Encoder rejects keys containing \`:\` with a clear error, OR decoder warns and drops them
- [ ] Test covering a map with \`:\`-key round-tripping through adapter → handler
- [ ] Dippin-lang parity: upstream also rejects — see \`export.flattenSteerContext\` in dippin-lang v0.22.0

## Source

PR #170 review round (security reviewer finding M2).

### LINKED ISSUE #172: test(adapter): cover CondOr / CondNot / Raw-wins / E2E .dip fixture for manager_loop
## Problem

Test coverage gaps surfaced during PR #170 review-squad pass:

1. **\`CondOr\` and \`CondNot\` formatting have zero coverage.** \`formatManagerLoopConditionExpr\` has explicit \`case ir.CondOr\` and \`case ir.CondNot\` branches. Only \`CondCompare\` and \`CondAnd\` are exercised. A typo in \`CondOr\` / \`CondNot\` case would silently produce empty \`stop_condition\` / \`steer_condition\` attrs.

2. **\`ir.Condition\` with both \`Raw\` and \`Parsed\` populated — no test for Raw-wins precedence.** \`managerLoopConditionText\` prefers \`Raw\`, but no test pins this. A future refactor that inverts preference to \`Parsed\` first would not be caught.

3. **No E2E adapter test using a real \`.dip\` fixture with \`manager_loop\`.** All existing tests use hand-constructed \`ir.Workflow\` structs. If dippin-lang v0.22.0 parses the \`manager_loop:\` stanza into different field names than expected, hand-constructed tests would pass while real \`.dip\` files would fail.

## Proposed change

Add three tests in \`pipeline/dippin_adapter_manager_loop_test.go\`:

- \`TestFromDippinIR_ManagerLoop_CondOrNotFormatting\` — covers the two missing condition-expression branches
- \`TestFromDippinIR_ManagerLoop_RawBeatsParsed\` — explicit assertion that \`Raw\` wins when both are populated
- \`TestDippinAdapter_E2E_ManagerLoop\` — loads a new \`pipeline/testdata/manager_loop.dip\` fixture through the real dippin-lang parser, then asserts the 6 flat attrs land correctly

Add \`pipeline/testdata/manager_loop.dip\` as the fixture.

## Acceptance criteria

- [ ] 3 new tests, ~100 LOC
- [ ] Coverage gap closes on all four adapter condition-formatter branches
- [ ] \`.dip\` E2E test doesn't depend on network — uses local testdata

## Source

PR #170 test-coverage reviewer findings.

### LINKED ISSUE #174: bug(adapter): nil ManagerLoopConfig silently produces missing subgraph_ref
## Problem

Surfaced during PR #170 review-squad pass. When an \`ir.Node\` has \`Kind: NodeManagerLoop\` but \`Config == nil\`, \`extractNodeAttrs\` returns nil and \`extractManagerLoopAttrs\` is never called. The adapter produces a valid graph node with no \`subgraph_ref\`; the failure surfaces late at \`Execute\` time with "missing subgraph_ref".

Dippin-lang's parser should always populate \`Config\` for a \`NodeManagerLoop\` node, so this is low-probability. Defensive adapter-level rejection would fail fast with better error messages.

## Proposed change

In \`pipeline/dippin_adapter.go\` (look for where \`extractNodeAttrs\` is called per kind), add a kind-specific guard:

\`\`\`go
if kind == ir.NodeManagerLoop && node.Config == nil {
    return nil, fmt.Errorf("adapter: manager_loop node %q has no ManagerLoopConfig", node.ID)
}
\`\`\`

Or equivalent. Other \`NodeKind\` variants may benefit from the same guard.

## Acceptance criteria

- [ ] Adapter returns a clear error at graph-build time, not at Execute time
- [ ] Test covering a hand-constructed IR with \`NodeManagerLoop\` and nil Config
- [ ] Consider applying the same pattern to other kinds that require non-nil Config

## Source

PR #170 test-coverage reviewer finding (gap 5).

### LINKED ISSUE #175: test(ci): add a manager_loop .dip example to exercise dippin doctor gate
## Problem

Surfaced during PR #170 review-squad pass. No example \`.dip\` in \`examples/\` uses a \`manager_loop\` node. The \`make doctor\` CI gate therefore gives no grade-A signal for the new node type. If a future dippin-lang release changes scoring or validation for \`manager_loop\` in a way that breaks tracker's adapter, CI won't catch it.

## Proposed change

Add a small example \`.dip\` that exercises \`manager_loop\`:

- Minimal child subgraph (can reuse an existing one or write a one-node child)
- Manager node with \`subgraph_ref\`, explicit \`poll_interval\`, \`max_cycles\`, and optionally \`stop_condition\` / \`steer_condition\` + \`steer_context\`
- Placed in \`examples/\` so it's picked up by the embedded workflows list and the doctor target

Verify \`dippin doctor examples/manager_loop_example.dip\` grades A.

## Acceptance criteria

- [ ] New example \`.dip\` in \`examples/\`
- [ ] \`make doctor\` includes it (either via automatic enumeration or Makefile update)
- [ ] Grade A
- [ ] \`tracker workflows\` lists it if it's embedded
- [ ] README section for built-in examples updated if embedded

## Source

PR #170 test-coverage reviewer finding (gap 10).

### LINKED ISSUE #176: polish(manager_loop): author-UX warnings + doc/comment accuracy cleanups
## Problem

Low-severity polish items surfaced during PR #170 review-squad pass. Bundled into one issue to reduce issue churn.

## Items

1. **Warn when both \`foo\` and \`manager.foo\` are set on the same node.** Today "unprefixed wins" silently. A pipeline author with both forms in a DOT file would benefit from a log warning pointing out that the legacy prefixed value was shadowed.

2. **Warn when \`stop_condition\` / \`steer_condition\` reference a context key that will never resolve.** Both conditions evaluate against the *parent* context, not the child's. An author writing \`stop_condition: "child_key = done"\` expecting to observe child-side state gets silent empty-string via \`resolveAndWarnVar\` with no actionable hint. A one-line diagnostic pointing at \`stack.child.*\` as the canonical observables would help.

3. **Fix inaccurate code comment in \`steerContextDecoder\`** (\`pipeline/handlers/manager_loop.go\`). The comment says "Sequences are listed longest-first so the replacer matches greedily" — \`strings.NewReplacer\` does single-pass left-to-right scans, not longest-match-greedy. The code works; the comment is misleading.

4. **Fix inaccurate test comment about sort order** (\`pipeline/dippin_adapter_manager_loop_test.go\`). Comment says "Sort order is by the ENCODED key (matches dippin-lang's exporter)" — actual code sorts by *original* (unencoded) key. The test happens to pass because for the chosen inputs both orderings produce the same result.

5. **\`TestManagerLoopHandler_EventsEmitted\` asserts only boolean flags.** No ordering or count checks. A regression emitting duplicate \`EventStageStarted\` events or firing \`EventStageCompleted\` before \`EventStageStarted\` wouldn't be caught. Upgrade to assert on slice of events in order.

6. **Asymmetric \`Parsed\` fallback in \`convertEdge\`** (pre-existing, newly conspicuous after PR #170). Edge conditions read \`irEdge.Condition.Raw\` with no fallback to a formatted \`Parsed\` tree. \`managerLoopConditionText\` in PR #170 *does* fall back. An edge with only \`.Parsed\` populated silently becomes unconditional.

## Acceptance criteria

Any subset is fine; items are independent. File a follow-up issue per item if work is split across people.

## Source

PR #170 review-squad pass (security L1/L2, correctness M4/L5, test coverage weak assertions).

clintecker:
## Summary

Bundles five follow-up fixes filed after the PR #170 review-squad pass on the manager_loop feature shipped in v0.23.0. One commit, tight scope, everything verified end-to-end.

### Closes #171 — Reject ':' in steer_context keys
- `flattenSteerContext` in `pipeline/dippin_adapter.go` now fails loudly at graph-build time when any key contains `:`. Rationale: dippin-lang v0.22.0's block-form formatter writes entries as `key: value` lines, so a colon in the key breaks `.dip`→IR→`.dip` round-trip. Upstream parser silently drops such keys with a diagnostic; tracker refuses to build the graph.
- Error: `ErrInvalidSteerContextKey` wrapping the offending key.
- Test: `TestFromDippinIR_ManagerLoop_RejectsColonInSteerContextKey`.

### Closes #172 — Test coverage gaps
Three new adapter tests:
- `TestFromDippinIR_ManagerLoop_CondOrNotFormatting` — asserts the exact textual output (`"outcome = success || status = done"`, `"not outcome = success"`) so a typo in the `||` or `not ` emitter fails a byte comparison instead of passing a semantic-only check.
- `TestFromDippinIR_ManagerLoop_RawBeatsParsed` — pins `managerLoopConditionText`'s preference: Raw wins over Parsed when both populated.
- `TestDippinAdapter_E2E_ManagerLoop` — loads `pipeline/testdata/manager_loop.dip` through the real `dippin-lang/parser` and asserts all 6 flat attrs (`subgraph_ref`, `poll_interval`, `max_cycles`, `stop_condition`, `steer_condition`, `steer_context`) land on the graph node.

### Closes #174 — Defensive nil-check for ManagerLoopConfig
- `convertNode` in `pipeline/dippin_adapter.go` now rejects a `NodeManagerLoop` with `Config == nil` at graph-build time. Before: silent pass-through, vague failure at Execute time. After: clear `ErrMissingManagerLoopCfg` naming the node.
- Scoped to `manager_loop` only per the issue — other kinds may get the same guard as follow-ups.
- Test: `TestFromDippinIR_ManagerLoop_NilConfigRejected`.

### Closes #175 — manager_loop example for CI doctor gate
- `examples/manager_loop_demo.dip` — small supervisor pipeline using `subgraph_ref`, `poll_interval`, `max_cycles`, `stop_condition`, `steer_condition`, `steer_context`. 25 lines.
- `examples/subgraphs/manager_loop_child.dip` — minimal child pipeline (DoWork → Done). 18 lines.
- Both grade A via `dippin doctor`.
- Added `manager_loop_demo.dip` to the `Makefile` doctor target so CI fails if a future change regresses adapter-path round-trip of the new attrs.

### Closes #176 — Bundled polish (6 items)
1. **Both-form warning in `managerAttr`.** When both `foo` and `manager.foo` are set on the same node, log a one-line diagnostic (`log.Printf`, matching the `[claude-code]`/`[webhook]` idiom in sibling handlers). Contract unchanged: unprefixed still wins.
2. **Unresolvable-key warning in `parseManagerLoopConfig`.** `warnUnknownStackChildKeys` scans `stop_condition` / `steer_condition` for `stack.child.<word>` references and warns on any subkey not in the canonical observables set (`stack.child.status`, `.cycles`, `.exit_status`). Narrow on purpose — bare keys like `outcome` or `status` are left alone because they are commonly set by parent pipelines.
3. **Misleading `steerContextDecoder` comment** — verification: the current comment (written in commit `dc566b9` during PR #170 round-4) does NOT claim "longest-first/greedy". It correctly says `strings.NewReplacer` "does not overlap matches" and "pattern order is not a correctness requirement" here. **No code change required.**
4. **Inaccurate sort-order test comment** — verification: the current comment in `TestFlattenSteerContext_EncodesReservedInKeysAndValues` already says "raw-key lexicographic order, not encoded-key order" (also fixed in `dc566b9`). **No code change required** beyond minor wording tidy to reflect the new colon-rejection semantics.
5. **`TestManagerLoopHandler_EventsEmitted` upgrade.** Boolean flags replaced with an ordered-slice assertion filtered to the manager's own `NodeID`. Checks: exactly one `EventStageStarted`, exactly one `EventStageCompleted`, ≥1 `EventManagerCycleTick` between them, started is first, completed is last. A regression firing completed before started or duplicating started now fails the test.
6. **Symmetric Parsed fallback in `convertEdge`.** `convertEdge` now reuses `managerLoopConditionText` (Raw-then-Parsed) rather than reading `.Raw` only. A Parsed-only `ir.Edge` (e.g. constructed by tests or simulate without running the parser) now produces a conditional edge instead of a silent unconditional one. Test: `TestConvertEdge_ParsedFallback`.

## Verification

- `go build ./...` — clean
- `go test ./... -short` — all 17 packages pass (3 consecutive runs green; pre-existing `TestManagerLoopHandler_CtxCancellation` timing flake confirmed to exist on `main` unchanged by this PR)
- `go test -race ./pipeline/... ./tui/...` — clean
- `gofmt -l .` — no output
- `go vet ./...` — clean
- `dippin doctor` on `ask_and_execute.dip`, `build_product.dip`, `build_product_with_superspec.dip`, `manager_loop_demo.dip` — all grade A
- `make doctor` — all grade A
- `make lint` — all `.dip` files pass lint

## Test plan

- [x] Adapter unit tests cover colon-in-key rejection, nil-config rejection, CondOr/CondNot formatting, Raw-beats-Parsed, E2E parse+adapt
- [x] Handler unit tests cover managerAttr both-form warning path (implicit via existing `TestParseManagerLoopConfig_UnprefixedWinsOverPrefixed` which prints the warning), event sequencing, Parsed-fallback for convertEdge
- [x] `make doctor` passes with new example
- [x] Race detector clean on `./pipeline/...` and `./tui/...`

## Summary by CodeRabbit

* **New Features**
  * Added example and E2E workflows demonstrating manager-loop supervision with polling, cycle limits, conditional stops, and steering hints.

* **Bug Fixes**
  * More explicit validation and clearer errors for missing or invalid manager-loop and steer-context configurations.

* **Improvements**
  * Runtime warnings for legacy/duplicate attributes and for unsupported child-subkey references; safer condition handling to prevent unintended edges.

* **Tests**
  * Strengthened unit and E2E tests with stricter event-order assertions and broader adapter coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ lint:
 
 doctor:
 	@FAIL=0; \
-	for f in examples/ask_and_execute.dip examples/build_product.dip examples/build_product_with_superspec.dip; do \
+	for f in examples/ask_and_execute.dip examples/build_product.dip examples/build_product_with_superspec.dip examples/manager_loop_demo.dip; do \
 		GRADE=$$($(DIPPIN) doctor "$$f" 2>&1 | grep 'Grade' | sed 's/.*Grade: //' | sed 's/  .*//'); \
 		SCORE=$$($(DIPPIN) doctor "$$f" 2>&1 | grep 'Score' | sed 's/.*Score: //' | sed 's/\/100//'); \
 		printf "%-50s %s  %s/100\n" "$$(basename $$f)" "$$GRADE" "$$SCORE"; \
--- a/examples/manager_loop_demo.dip
+++ b/examples/manager_loop_demo.dip
@@ -12,7 +12,7 @@ workflow ManagerLoopDemo
     subgraph_ref: subgraphs/manager_loop_child.dip
     poll_interval: 30s
     max_cycles: 12
-    stop_condition: stack.child.status = success
+    stop_condition: stack.child.cycles = 10
     steer_condition: stack.child.cycles = 5
     steer_context:
       hint: halfway_through
--- a/examples/manager_loop_demo.dip
+++ b/examples/manager_loop_demo.dip
@@ -1,0 +1,31 @@
+workflow ManagerLoopDemo
+  goal: "Showcase the manager_loop supervisor: launch a child pipeline, poll for completion, steer mid-flight, and stop on a condition."
+  start: Supervise
+  exit: Done
+
+  defaults
+    model: claude-sonnet-4-6
+    provider: anthropic
+
+  manager_loop Supervise
+    label: "Supervise child pipeline"
+    subgraph_ref: subgraphs/manager_loop_child.dip
+    poll_interval: 30s
+    max_cycles: 12
+    stop_condition: stack.child.status = success
+    steer_condition: stack.child.cycles = 5
+    steer_context:
+      hint: halfway_through
+      priority: high
+
+  agent Done
+    label: "Summarize the supervised run."
+    prompt:
+      The manager loop completed.
+      Final child status: ${ctx.stack.child.status}
+      Cycles elapsed: ${ctx.stack.child.cycles}
+
+      Summarize the outcome in one paragraph.
+
+  edges
+    Supervise -> Done
--- a/examples/subgraphs/manager_loop_child.dip
+++ b/examples/subgraphs/manager_loop_child.dip
@@ -1,0 +1,23 @@
+workflow ManagerLoopChild
+  goal: "Child pipeline supervised by manager_loop_demo.dip — performs one cycle of work per run."
+  start: DoWork
+  exit: Done
+
+  agent DoWork
+    label: "Perform one unit of supervised work."
+    auto_status: true
+    prompt:
+      You are a worker supervised by a manager_loop. Perform one cycle of
+      the assigned task. If the steering hint ${ctx.hint} is set, use it
+      to bias your approach.
+
+      End your response with STATUS: success when the work is complete,
+      or STATUS: fail if another cycle is required.
+
+  agent Done
+    label: "Finalize cycle result."
+    prompt:
+      Briefly summarize this cycle's outcome for the manager.
+
+  edges
+    DoWork -> Done
--- a/pipeline/dippin_adapter.go
+++ b/pipeline/dippin_adapter.go
@@ -14,11 +14,13 @@ import (
 )
 
 var (
-	ErrNilWorkflow     = errors.New("nil workflow")
-	ErrMissingStart    = errors.New("workflow missing Start node")
-	ErrMissingExit     = errors.New("workflow missing Exit node")
-	ErrUnknownNodeKind = errors.New("unknown node kind")
-	ErrUnknownConfig   = errors.New("unknown config type")
+	ErrNilWorkflow            = errors.New("nil workflow")
+	ErrMissingStart           = errors.New("workflow missing Start node")
+	ErrMissingExit            = errors.New("workflow missing Exit node")
+	ErrUnknownNodeKind        = errors.New("unknown node kind")
+	ErrUnknownConfig          = errors.New("unknown config type")
+	ErrInvalidSteerContextKey = errors.New("steer_context key contains ':' which breaks block-form round-trip through the .dip formatter")
+	ErrMissingManagerLoopCfg  = errors.New("manager_loop node is missing required ir.ManagerLoopConfig")
 )
 
 // FromDippinIR converts a Dippin IR Workflow to a Tracker Graph.
@@ -141,6 +143,16 @@ func convertNode(irNode *ir.Node) (*Node, error) {
 		return nil, fmt.Errorf("%s: %w", irNode.Kind, ErrUnknownNodeKind)
 	}
 
+	// Kind-specific required-config guard. A manager_loop node with a nil
+	// Config would otherwise flow through extractNodeAttrs as a no-op and
+	// produce a graph node without subgraph_ref, surfacing later at Execute
+	// time as a vague runtime error. Fail loudly at build time instead.
+	// Scoped to manager_loop for now — we may extend the same guard to other
+	// kinds as follow-ups if they exhibit the same silent-degrade pattern.
+	if irNode.Kind == ir.NodeManagerLoop && irNode.Config == nil {
+		return nil, fmt.Errorf("node %s: %w", irNode.ID, ErrMissingManagerLoopCfg)
+	}
+
 	gNode := &Node{
 		ID:    irNode.ID,
 		Shape: shape,
@@ -194,7 +206,7 @@ func extractValueNodeAttrs(config ir.NodeConfig, attrs map[string]string) (bool,
 	case ir.ConditionalConfig:
 		// Conditional nodes are pure routing — no config to extract.
 	case ir.ManagerLoopConfig:
-		extractManagerLoopAttrs(cfg, attrs)
+		return true, extractManagerLoopAttrs(cfg, attrs)
 	default:
 		return false, nil
 	}
@@ -453,7 +465,7 @@ func extractSubgraphAttrs(cfg ir.SubgraphConfig, attrs map[string]string) {
 // tracker defaults. If/when the handler grows those modes this extractor is
 // the right place to translate the zero-value IR sentinels into the
 // corresponding handler attrs.
-func extractManagerLoopAttrs(cfg ir.ManagerLoopConfig, attrs map[string]string) {
+func extractManagerLoopAttrs(cfg ir.ManagerLoopConfig, attrs map[string]string) error {
 	if cfg.SubgraphRef != "" {
 		attrs["subgraph_ref"] = cfg.SubgraphRef
 	}
@@ -469,9 +481,14 @@ func extractManagerLoopAttrs(cfg ir.ManagerLoopConfig, attrs map[string]string) 
 	if s := managerLoopConditionText(cfg.SteerCondition); s != "" {
 		attrs["steer_condition"] = s
 	}
-	if s := flattenSteerContext(cfg.SteerContext); s != "" {
+	s, err := flattenSteerContext(cfg.SteerContext)
+	if err != nil {
+		return err
+	}
+	if s != "" {
 		attrs["steer_context"] = s
 	}
+	return nil
 }
 
 // managerLoopConditionText extracts the best textual form of a condition:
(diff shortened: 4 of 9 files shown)


clintecker:
## Round-4 review fixes — 7/7 findings addressed

All seven findings from CodeRabbit / Copilot / Codex are addressed across two commits:

**Commit `80f6300` — adapter behavior (findings #1, #2)**

- **#1 (Codex P2)**: `convertEdge`'s Parsed fallback no longer emits paren-bearing expressions that crash the edge evaluator at runtime. When a Condition has only `.Parsed` populated and the formatted text contains `(` or `)`, `convertEdge` now returns `ErrParenthesizedParsedCondition` with a clear remediation hint. Flat all-AND / all-OR Parsed trees continue to work (no parens). Raw-populated conditions are trusted as-is. Signature change: `convertEdge(*ir.Edge) *Edge` → `convertEdge(*ir.Edge) (*Edge, error)`, propagated through `addIREdges` and `FromDippinIR`. Three new tests: flat happy path, mixed-precedence rejection, and Raw-bypass-with-parens.
- **#2 (Copilot)**: `convertNode` returned `fmt.Errorf("node %s: %w", id, ErrMissingManagerLoopCfg)`; `addIRNodes` then wrapped again, producing `node mgr: node mgr: ...`. Return the sentinel bare; the outer wrapper adds the single node-context prefix. Existing `TestFromDippinIR_ManagerLoop_NilConfigRejected` still passes because its string check looks for `"mgr"`, which the outer wrapper provides.

**Commit `fd73986` — documentation & fixture cleanup (findings #3–#7)**

- **#3**: `pipeline/handlers/manager_loop.go` comment now references `extractStackChildKeys` (the real function) instead of the non-existent `stackChildKeyPattern`.
- **#4**: `pipeline/dippin_adapter_manager_loop_test.go` comment now references `TestFromDippinIR_ManagerLoop_RejectsColonInSteerContextKey` (the real test) instead of `TestFlattenSteerContext_RejectsColonInKey`.
- **#5**: `pipeline/testdata/manager_loop.dip` changed `stop_condition: stack.child.outcome = success` to `stack.child.status = success`. The handler never writes `.outcome` — only `.status`, `.cycles`, `.exit_status`. This also removes a false unknown-observable warning against the adapter's own fixture.
- **#6**: `TestDippinAdapter_E2E_ManagerLoop` expected-value table updated in lockstep to match the fixture change.
- **#7**: `examples/manager_loop_demo.dip` changed `stop_condition: stack.child.status = success` (unreachable — `status=success` only fires on child completion, at which point the handler returns before stop_condition is re-evaluated) to `stack.child.cycles = 10`. With `max_cycles: 12`, this actually triggers during the loop, giving the example a demoable stop.

## Verification

```
go build ./...              # clean
go test ./... -short        # all 17 packages green
gofmt -l .                  # empty
go vet ./...                # clean
make doctor                 # all 4 examples grade A (via dippin v0.22.0)
```

Commit SHAs: `80f6300` and `fd73986`.

clintecker:
## Round 3 Copilot review — responses

Commit: 50caf76

### 1. Warning log leaks attribute values (manager_loop.go `managerAttr`)
Fixed. Changed the `both forms set` diagnostic to log only the attribute keys — never the values — so a future use of manager_loop attrs that happens to carry sensitive material (tokens in URLs, paths, etc.) cannot leak into logs. Kept the "unprefixed wins, delete legacy" guidance unchanged.

### 2. Orphan doc comment block above `warnUnknownStackChildKeys`
Fixed. Deleted the stale `extractStackChildKeys …` comment block that was sitting above `warnUnknownStackChildKeys`. The real doc comment for `extractStackChildKeys` already lives on the function itself.

### 3. `strings.SplitSeq` Go-version concern
Rejected. `go.mod` is on `go 1.25.5`, and `strings.SplitSeq` is available from Go 1.24. No change needed.

### 4. `warnUnknownStackChildKeys` does not identify which node owns the warning
Fixed. Threaded the owning node's ID through `parseManagerLoopConfig(nodeID, attrs)` and `warnUnknownStackChildKeys(nodeID, attrName, expr)`. The warning line now reads:
> `[manager_loop] warning: node "<id>" <attr> references "<key>" which is not a known observable; known keys: stack.child.status, stack.child.cycles, stack.child.exit_status`

All test call sites updated to pass a `"Supervise"` sentinel node ID.

### 5. Test name says `_WithParens` but Raw had no parens (`TestConvertEdge_RawWinsOverParsed_WithParens`)
Fixed. Raw is now `"(ctx.a = 1 || ctx.b = 2) && ctx.c = 3"`, and the test asserts the returned `Edge.Condition` equals that string verbatim. This makes the test genuinely prove that Raw bypasses paren rejection.

### 6. Comment misrepresents evaluator failure mode (`TestConvertEdge_ParsedFallback_MixedPrecedenceRejected`)
Fixed. Replaced the "invalid variable name `(b`" wording with the real behavior: `EvaluateCondition` treats tokens like `(b` as unknown variables, logs a warning, resolves them to `""`, and the clause may then evaluate incorrectly.

### 7. `ErrParenthesizedParsedCondition` doc describes wrong failure mode
Fixed. Rewrote the doc to match what the pipeline evaluator actually does: unknown-variable warning + empty-string resolution + silent mis-evaluation. Kept the guidance that authors should populate `Condition.Raw` with a flat form or simplify the `Parsed` tree.

## Verification

- `go build ./...` — pass
- `go test ./... -short` — all 17 packages pass
- `gofmt -l .` — clean
- `go vet ./...` — clean
- `make doctor` — `ask_and_execute.dip` A 95/100, the other three A 100/100

PR #198, opened 2026-05-04
Add dispatch_sprints + TerminalTool + 4-strategy SR matcher (architect-side machinery for local codegen)

sugi-chan:
## Summary

Tracker-side machinery for the local-codegen sprint-authoring pipeline in [`2389-research/pipelines#2`](https://github.com/2389-research/pipelines/pull/2). Lands the **structural fix** that turns per-sprint dispatch from an LLM agent loop (prone to self-revision after the last sprint completed) into a deterministic in-tool loop with retry + partial-apply + tolerant parsing.

### New tool: `dispatch_sprints`
Reads a JSONL plan (one `{path, description}` per line) and runs the per-sprint author+audit pipeline once per line. Implements `TerminalTool` so the agent session ends immediately on success — no more wasted post-dispatch turns. Per-sprint failures get bounded retry-with-backoff for transient provider errors (5xx / rate limit / timeout / network), classified via the existing `llm.ProviderErrorInterface.Retryable()`. Non-retryable errors (auth, invalid request) bubble back to the agent so it can fix the JSONL and retry.

### New runtime primitive: `TerminalTool` interface
Tools may opt in via `IsTerminal() bool`. Agent runtime checks the flag after each tool invocation: if a terminal tool succeeds, the loop breaks immediately. Errors keep the loop alive (so the model can react and retry). 3-line change in `agent/session.go` + a returns-bool refactor of `executeToolCalls`.

### `write_enriched_sprint` overhauled
- **Refactor:** extracted per-sprint author+audit logic into a public `RunOne()` so `dispatch_sprints` can call it without going through the LLM tool dispatch path.
- **MaxTokens 16384** on both author and audit passes — the prior default (~4-8K) was truncating foundation sprints. NIFB sprint 001 hit this with 18 ORM models + auth + conftest + tests, ran out of room mid-Rules section.
- **4-strategy SR-block matcher:** exact → indent-preserving → whitespace-insensitive → fuzzy (Levenshtein ratio ≥0.9). Mirrors the `lib/merge_sr.py` reference used by the runner side.
- **Partial-apply semantics:** a block N failing no longer poisons blocks 1..N-1. The new `PATCHED-PARTIAL` verdict distinguishes \"some blocks landed, some skipped\" from clean `PATCHED`. Reduces `PASS-FALLBACK-NOMATCH` rate from 31% to ~5% on NIFB-scale runs.
- **Tolerant audit verdict parser:** finds `AUDIT-VERDICT:` anywhere in the first 10 non-empty lines (not just literal first line); strips enclosing markdown fences; handles inline backtick/bold/italic decoration; tolerates leading prose like \"Looking at the draft carefully...\".
- **System prompt rewritten** to use the validated speedrun section list (Tricky semantics, Data contract, API contract, Algorithm, Test contract, Verbatim files) instead of the older shape. Adds Pattern A vs Pattern B decision criteria + non-overlap discipline for the auditor.
- **Few-shot example replaced** with NIFB v7 SPRINT-001 (foundation, 828 lines) + SPRINT-002 (additive, 222 lines). Sonnet writer sees both shapes as anchors.

### Registration
`pipeline/handlers/backend_native.go` registers `dispatch_sprints` alongside `write_enriched_sprint` when `TRACKER_SPRINT_WRITER_MODEL` is set. They share the same `WriteEnrichedSprintTool` instance so per-sprint author+audit logic isn't duplicated.

## Tests

**32 new unit tests across 2 files**, full tracker suite (12 packages) green.

`agent/tools/dispatch_sprints_test.go` (7 tests):
- `TestDispatchSprintsHappyPath` — 3 sprints, all pass, 6 LLM calls
- `TestDispatchSprintsRejectsInvalidJSONL` — 5 sub-cases (malformed JSON, missing path, invalid path format, empty description, empty file)
- `TestDispatchSprintsContinuesOnFailure` — non-strict mode keeps going after a per-sprint failure
- `TestDispatchSprintsStrictHaltsOnFailure` — strict mode aborts on first failure
- `TestRunOneWithRetry_TransientThenSucceeds` — RateLimitError + ServerError retried; third attempt's success returned
- `TestRunOneWithRetry_NonRetryableErrorAborts` — AuthenticationError fails immediately, no retry
- `TestExecute_PatchedPartial` — audit emits 2 SR blocks (1 matches, 1 doesn't); verdict = PATCHED-PARTIAL with patches=1; partial fix on disk

`agent/tools/write_enriched_sprint_test.go` (25 tests):
- All 4 SR match strategies + cross-strategy ordering invariant (`exact` wins over `fuzzy`)
- Partial-apply: applies block 0 + block 2, skips block 1 cleanly
- Tolerant verdict parser: 7 format variations (markdown fence, leading preamble, backtick/bold/italic decoration, trailing punctuation, multiple blank lines, prose-then-verdict, blank line gap)
- Levenshtein similarity ratio (6 known pairs) + common-leading-indent helper (7 cases)
- Multi-block ordering, empty-search rejection, fuzzy threshold rejection

## Validated end-to-end via the pipelines repo

| Project | Stage | Outcome | Cost | Time |
|---|---|---|---|---|
| Notebook synthetic (3 sprints) | Architect + Runner | 41/41 pytest passing | ~\$2 | 28 min (May 4 v5) |
| Notebook synthetic | Architect + Runner | 34/34 pytest passing | ~\$2.30 | 17 min (May 1 v4) |
| NIFB (16 sprints) | Architect-only | All 16 sprints, Pattern B chosen autonomously, validate_output green | ~\$5 | 47 min (May 1 v3) |

vs all-cloud equivalent (Opus architect + gpt-5.4 codegen): roughly **4× more expensive (~\$8-9)** for the same Notebook run. Local-first wins by routing per-file work to qwen.

## Test plan

- [x] `go test ./...` — 12 packages green (cmd/tracker-conformance, llm + 4 provider variants, pipeline + handlers, tui + render, agent + tools)
- [x] `go build ./...` clean
- [x] Smoke-tested via Notebook synthetic + NIFB architect-only (see runs above)
- [ ] Stress test with the parallel-mode dispatch_sprints variant (out of scope; future work)
- [ ] qwen runner end-to-end on NIFB scale (16 backend sprints + frontend) — needs runner-side stack-detection improvements not in this PR

## Companion PR

**[`2389-research/pipelines#2`](https://github.com/2389-research/pipelines/pull/2)** — lands the dip + docs side. Together they're a self-contained local-codegen pipeline ready for use.

## Summary by CodeRabbit

* **New Features**
  * Sessions now stop early when a successful "terminal" tool runs.
  * Code generation: create single- or multi-file outputs from a contract.
  * Sprint enrichment: produce audited sprint Markdown with SEARCH/REPLACE patching and partial-patch handling.
  * Batch sprint dispatch: run multiple sprints with per-sprint retries, summaries, strict/non-strict modes, and failure reporting.
  * Enable codegen and sprint workflows via environment variables.

* **Tests**
  * Extensive tests for sprint writer, dispatcher, retry behavior, SR-block matching, and audit parsing.

* **Documentation**
  * Added a comprehensive sprint example and contract guidance.

### DIFF
--- a/agent/session.go
+++ b/agent/session.go
@@ -259,9 +259,9 @@ func (s *Session) executeTurn(ctx context.Context, turn int, start time.Time, tr
 		return done, stop, err
 	}
 
-	stopped := s.handleToolCalls(ctx, toolCalls, resp, turn, turnStart, tracker, prevCacheHits, prevCacheMisses, result, ts)
-	if stopped {
-		return false, true, nil
+	stop, naturally := s.handleToolCalls(ctx, toolCalls, resp, turn, turnStart, tracker, prevCacheHits, prevCacheMisses, result, ts)
+	if stop {
+		return naturally, true, nil
 	}
 
 	// Run the verify-after-edit loop if any edit tools were called this turn.
@@ -309,8 +309,11 @@ func (s *Session) handleNoTools(resp *llm.Response, turn int, turnStart time.Tim
 }
 
 // handleToolCalls processes a turn where the LLM returned tool calls.
-// Returns true if the loop should stop (loop detected).
-func (s *Session) handleToolCalls(ctx context.Context, toolCalls []llm.ToolCallData, resp *llm.Response, turn int, turnStart time.Time, tracker *ContextWindowTracker, prevCacheHits, prevCacheMisses int, result *SessionResult, ts *turnState) bool {
+// Returns:
+//   - stop: the outer loop should break.
+//   - naturally: the stop was a clean end (terminal-tool success), as opposed
+//     to a loop-detection abort. Only meaningful when stop=true.
+func (s *Session) handleToolCalls(ctx context.Context, toolCalls []llm.ToolCallData, resp *llm.Response, turn int, turnStart time.Time, tracker *ContextWindowTracker, prevCacheHits, prevCacheMisses int, result *SessionResult, ts *turnState) (stop, naturally bool) {
 	signature := s.computeToolSignature(toolCalls)
 	if signature == ts.lastToolSignature {
 		ts.consecutiveLoopCount++
@@ -325,14 +328,14 @@ func (s *Session) handleToolCalls(ctx context.Context, toolCalls []llm.ToolCallD
 		result.LoopDetected = true
 		s.emitTurnMetrics(turn, turnStart, resp, tracker, prevCacheHits, prevCacheMisses, result)
 		s.emit(Event{Type: EventTurnEnd, SessionID: s.id, Turn: turn})
-		return true
+		return true, false
 	}
 
-	hadErrors := s.executeToolCalls(ctx, toolCalls, result)
+	hadErrors, terminate := s.executeToolCalls(ctx, toolCalls, result)
 	s.maybeInjectReflection(hadErrors, ts)
 	s.emitTurnMetrics(turn, turnStart, resp, tracker, prevCacheHits, prevCacheMisses, result)
 	s.emit(Event{Type: EventTurnEnd, SessionID: s.id, Turn: turn})
-	return false
+	return terminate, terminate
 }
 
 // maybeInjectReflection appends the structured reflection prompt when tool errors
@@ -467,7 +470,7 @@ func (s *Session) runRepairTurn(ctx context.Context, result *SessionResult) erro
 		return nil // LLM responded with text only (e.g. "I fixed it")
 	}
 
-	s.executeToolCalls(ctx, toolCalls, result)
+	_, _ = s.executeToolCalls(ctx, toolCalls, result)
 	return nil
 }
 
--- a/agent/session.go
+++ b/agent/session.go
@@ -14,10 +14,10 @@ import (
 	"github.com/2389-research/tracker/llm"
 )
 
-// Completer is the interface needed from the LLM client.
-type Completer interface {
-	Complete(ctx context.Context, req *llm.Request) (*llm.Response, error)
-}
+// Completer is the interface needed from the LLM client. It's an alias of
+// tools.Completer so both packages refer to the same type — preventing
+// silent divergence if either side grows new methods.
+type Completer = tools.Completer
 
 // SessionOption configures a Session.
 type SessionOption func(*Session)
--- a/agent/session_run.go
+++ b/agent/session_run.go
@@ -261,18 +261,30 @@ func (s *Session) computeToolSignature(toolCalls []llm.ToolCallData) string {
 	return strings.Join(toolSigs, ",")
 }
 
-// executeToolCalls runs each tool call sequentially and appends results to messages.
-// It returns true if any tool call resulted in an error.
-func (s *Session) executeToolCalls(ctx context.Context, toolCalls []llm.ToolCallData, result *SessionResult) bool {
+// executeToolCalls runs each tool call sequentially and appends results to
+// messages. Returns:
+//   - hadErrors: any tool call in the batch errored (used to drive reflection).
+//   - terminate: a tool implementing TerminalTool succeeded; the caller should
+//     break out of the agent loop.
+//
+// On terminal-tool success the loop short-circuits: later tool calls in the
+// same batch are skipped. Running them would feed results to a session that
+// the model never gets a chance to react to.
+func (s *Session) executeToolCalls(ctx context.Context, toolCalls []llm.ToolCallData, result *SessionResult) (hadErrors, terminate bool) {
 	var toolResults []llm.ContentPart
-	anyError := false
 	for _, call := range toolCalls {
 		toolResult, toolDuration := s.executeSingleTool(ctx, call)
 		result.ToolCalls[call.Name]++
 		if toolResult.IsError {
-			anyError = true
+			hadErrors = true
 		}
 		s.episodeLog.Record(call.Name, string(call.Arguments), toolResult.Content, toolResult.IsError)
+
+		if !toolResult.IsError {
+			if tool := s.registry.Get(call.Name); tool != nil && tools.IsToolTerminal(tool) {
+				terminate = true
+			}
+		}
 
 		s.emit(Event{
 			Type:         EventToolCallEnd,
@@ -287,13 +299,17 @@ func (s *Session) executeToolCalls(ctx context.Context, toolCalls []llm.ToolCall
 			Kind:       llm.KindToolResult,
 			ToolResult: &toolResult,
 		})
+
+		if terminate {
+			break
+		}
 	}
 
 	s.messages = append(s.messages, llm.Message{
 		Role:    llm.RoleTool,
 		Content: toolResults,
 	})
-	return anyError
+	return hadErrors, terminate
 }
 
 // executeSingleTool runs a single tool call, handling caching.
--- a/agent/session_run.go
+++ b/agent/session_run.go
@@ -267,9 +267,12 @@ func (s *Session) computeToolSignature(toolCalls []llm.ToolCallData) string {
 //   - terminate: a tool implementing TerminalTool succeeded; the caller should
 //     break out of the agent loop.
 //
-// On terminal-tool success the loop short-circuits: later tool calls in the
-// same batch are skipped. Running them would feed results to a session that
-// the model never gets a chance to react to.
+// On terminal-tool success the loop also breaks immediately, skipping any
+// remaining tool calls in the same LLM response. This is intentional: once
+// a terminal tool has done its job, subsequent calls in the same turn are
+// either redundant or could perform unwanted side-effects after the "final
+// step" (e.g., a model emitting `[dispatch_sprints, write_summary_doc]`
+// shouldn't run write_summary_doc once dispatch_sprints succeeded).
 func (s *Session) executeToolCalls(ctx context.Context, toolCalls []llm.ToolCallData, result *SessionResult) (hadErrors, terminate bool) {
 	var toolResults []llm.ContentPart
 	for _, call := range toolCalls {
@@ -280,9 +283,14 @@ func (s *Session) executeToolCalls(ctx context.Context, toolCalls []llm.ToolCall
 		}
 		s.episodeLog.Record(call.Name, string(call.Arguments), toolResult.Content, toolResult.IsError)
 
+		// Terminal-tool check: a successful invocation of a tool that
+		// flags itself terminal ends the agent session. Errors keep the
+		// loop alive so the model can react to the failure.
+		thisIsTerminal := false
 		if !toolResult.IsError {
 			if tool := s.registry.Get(call.Name); tool != nil && tools.IsToolTerminal(tool) {
 				terminate = true
+				thisIsTerminal = true
 			}
 		}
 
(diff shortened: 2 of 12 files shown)


clintecker (changes requested):
# Consolidated review — combines my findings with Copilot's and CodeRabbit's

The architecture is sound (deterministic in-tool dispatch instead of an LLM-loop, partial-apply, retry-on-transient) and the test coverage is substantial. There are a handful of issues to address before this merges, the most important being a **wrong-position replace bug** that affects three of the four SR-match strategies.

The two earlier rounds of fixes (`a80539c` for I/O error surfacing, `81998ef` for nil guards) addressed the corresponding CodeRabbit findings — those are not repeated here.

---

## Blocking

**B1. `strings.Replace(content, chunk, …, 1)` patches the wrong location** *(Copilot, write_enriched_sprint.go:912)*
The whitespace/indent/fuzzy matchers locate a specific N-line window by index, but then apply the replacement via global `strings.Replace` of the chunk text. If the same chunk appears earlier in the document, the *first* occurrence is replaced — not the matched window. Apply by byte offset (`content[:start] + replacement + content[end:]`) using the indices the matcher already computed. This affects three of the four strategies; it's a real correctness bug, not a fuzzy-only edge case.

**B2. Path traversal in `RunOne` / `Execute` (sprint paths)** *(Copilot + me, write_enriched_sprint.go:644)*
`filepath.Join(outputDir, path)` discards `outputDir` if `path` is absolute, and accepts `../` segments. The `dispatch_sprints` caller validates against `^SPRINT-\d{3}\.md$`, so it's safe — but a direct `Execute` call from the LLM tool path has no such guard. Mirror the existing `safePatchPath` pattern: reject absolute paths, `filepath.Clean`, and verify the resolved path stays under `outputDir`.

**B3. Path traversal in `generate_code.go`** *(Copilot, generate_code.go:166)*
Same issue, both for sequential `f.Path` and the `# === FILE: ... ===` headers parsed from model output. Apply the same guard.

**B4. `LoadContract` allows abs/.. paths and ships file contents to the LLM** *(Copilot, write_enriched_sprint.go:451)*
`contract_file` is read and forwarded to the provider as prompt context. With no path guard, a confused or adversarial model can exfiltrate arbitrary files (e.g. `~/.aws/credentials`). Restrict to `workDir` (or a dedicated `.ai/` subtree).

**B5. `TerminalTool` doesn't actually terminate immediately** *(Copilot + me, session_run.go:195)*
`terminate=true` is set on terminal-tool success, but the surrounding loop continues processing the rest of the tool-call batch. This means (a) any side-effecting tools after the "final" one still run, and (b) `terminate` stays true even if a later call errored. Break out of the tool-call loop the moment a terminal tool succeeds.

## Should fix

**S1. Audit pass silently soft-fails on provider error** *(me, write_enriched_sprint.go:536-539)*
On audit-pass error the code logs to stderr, sets `auditResp = nil`, and ships the draft as `verdict = PASS` — indistinguishable from a real auditor pass. This is a direct violation of CLAUDE.md's "provider errors must hard-fail." The author pass at line 501 propagates correctly; mirror it. (At minimum, emit a distinct `PASS-FALLBACK-AUDITFAIL` verdict so callers can tell.)

**S2. `isRetryableError` should use `errors.As()`** *(CodeRabbit, dispatch_sprints.go:266-272)*
The manual unwrap loop works for current call sites but `errors.As()` is the idiomatic way and correctly handles joined-error chains.

**S3. `auditUserPrompt` contradicts `auditSystemPrompt`** *(me)*
User prompt instructs "second line must be the `# Sprint NNN — Title (enriched spec)` heading"; system prompt says PASS responses are "ONLY this single line." Models that follow the user prompt waste tokens; the parser silently drops the extra markdown. Reconcile to the system-prompt rules.

**S4. Duplicate `Completer` interface** *(CodeRabbit + me)*
Declared in `agent/session.go:17-20`, `agent/tools/generate_code.go:17`, and `agent/tools/write_enriched_sprint.go`. Identical method sets so it compiles, but any future divergence will silently break only one call site. Consolidate into `agent/tools/completer.go` (or pull from `agent.Completer`).

**S5. `generate_code.go` default model `"gpt-5.4-nano"` isn't a real model** *(me, generate_code.go:51)*
Env-gating in `backend_native.go` requires `TRACKER_CODEGEN_MODEL` so the override always wins, but a caller constructing the tool without the option will hard-fail with model-not-found. Use `gpt-4o-mini` or remove the default.

## Nits

**N1. `TestRunOneWithRetry_TransientThenSucceeds` nil-fallback panic risk** *(me, dispatch_sprints_test.go:354)*
`mock.fallback` is nil; if the call count is off by one, the test panics rather than fails cleanly. Set a sentinel that returns a non-nil error.

---

I'm going to take this on directly: rebase the branch (it's currently 430 commits behind main, so there's real conflict surface), apply the fixes above, and re-run the suite. Will follow up here once that's pushed.

sugi-chan:
@clintecker — thanks for the consolidated review. Just a heads-up: I pushed a fix-pass commit (`ee23412`) ~3 min before your review landed, which already addresses **B1, B2, B4, B5** from your blocking list:

- **B1** (SR matcher position bug) — added `replaceLineWindow()` that replaces by byte offset using the indices the windowed strategies already compute; all three of whitespace/indent/fuzzy now route through it. Exact-match path keeps `strings.Replace` since first-occurrence is intentional there.
- **B2** (path traversal in `RunOne`) — abs-path reject + `..` reject + `filepath.Rel(outputDir, resolved)` check that `rel` doesn't start with `..`.
- **B4** (LoadContract exfiltration vector) — new `resolveUnderWorkDir()` helper that returns the abs resolved path and validates it stays under `workDir`. Reused for both contract_file and (in essence) the sprint output path. Allows abs paths whose resolved form is under workDir (needed for `t.TempDir()` based tests); rejects everything else.
- **B5** (TerminalTool early break) — the per-call loop in `executeToolCalls` now breaks immediately after a terminal tool succeeds, after appending its result so the message stays coherent.

**Still outstanding from your blocking list:** B3 (`generate_code.go` path traversal) — I'd explicitly scoped it out as preexisting code on this branch (not introduced by this PR), but happy to be told otherwise.

**Untouched from your Should-fix:** S1 (audit-fail soft-pass should be a distinct `PASS-FALLBACK-AUDITFAIL` verdict — good catch, didn't think about that one), S2 (errors.As idiom), S3 (audit prompt user/system contradiction), S4 (duplicate Completer interface), S5 (gpt-5.4-nano default), N1 (test nil-fallback).

If you're taking the rebase + remaining fixes, I'll stay out of your way. Let me know if anything in `ee23412` doesn't match what you had in mind and I'll revert.

clintecker:
Status update following [my earlier review](https://github.com/2389-research/tracker/pull/198#pullrequestreview-): I rebased the branch onto current main (was 430 commits behind) and pushed two new commits.

Sugi-chan's `ee23412` (pr-feedback-3) had already addressed:
- SR matcher position bug (replaced `strings.Replace(content, chunk, ...)` with byte-range `replaceLineWindow`)
- Path traversal in `LoadContract` and `RunOne` (new `resolveUnderWorkDir` helper)
- Terminal-tool early break on success

That work was preserved verbatim through the rebase — only the conflict resolution in `agent/session.go` and `agent/session_run.go` changed (main has refactored the turn loop into helpers; merged in main's `hadErrors` return alongside sugi-chan's `terminate` so reflection-on-error still works).

Then `5e28345` (pr-feedback-4) covers the items the earlier round didn't:

**Correctness**
- Audit pass no longer soft-fails on provider errors (`write_enriched_sprint.go:601`). Was logging to stderr and shipping `verdict=PASS` — direct CLAUDE.md violation. Now propagates the error like the author pass.
- `auditUserPrompt` was contradicting `auditSystemPrompt` (telling the model to write a sprint heading after the verdict, even on PASS). Rewrote.

**Security**
- `generate_code.go` path traversal: three sites (`generateSequential`'s `f.Path`, `generateSingleCall`'s default, and the `# === FILE: ===`-parsed names) now go through a shared `resolveUnderRoot` helper. Sugi-chan's `resolveUnderWorkDir` is a method bound to `WriteEnrichedSprintTool`; the new helper is a package-level function so `generate_code.go` can use it without coupling.

**Quality**
- `isRetryableError` uses `errors.As()` instead of manual unwrap.
- `Completer` interface consolidated: `tools.Completer` is now the single canonical declaration; `agent.Completer` is a type alias of it (using the existing one-way `agent → tools` import). No more parallel declarations to drift apart.
- `gpt-5.4-nano` default → `gpt-4o-mini` (the env-gated registration always overrides, but the default itself was misleading).
- `retryMockCompleter` returns a clean error on script overrun instead of nil-deref panicking.

Build clean, all 17 packages green under `go test ./... -short`. Force-pushed with `--force-with-lease` anchored to `ee23412`.

clintecker:
Pushed `186ce47` (pr-feedback-5) — addresses CodeRabbit's latest review and the Quality Gates CI fail.

**CI:** the failure was a gofmt nit (`dispatch_sprints.go` and `write_enriched_sprint_test.go` had var-block alignment drift). Fixed.

**CodeRabbit findings (all 6):**

| # | File:line | Fix |
|---|-----------|-----|
| 1 | `completer.go:50` — `resolveUnderRoot` symlink escape | Added `filepath.EvalSymlinks` on both root and candidate before containment check; new `evalSymlinksLeniently` handles write-paths whose target doesn't yet exist by resolving the longest existing ancestor. 5 new tests in `completer_test.go`. |
| 2 | `write_enriched_sprint.go:487` — `LoadContract` symlink bypass | `resolveUnderWorkDir` now delegates to the shared `resolveUnderRoot`, inheriting the symlink protection. |
| 3 | `write_enriched_sprint.go:501` — `resolveOutputDir` doesn't clamp absolute paths | Now confines `output_dir` under `workDir`. Returns an error when escape is attempted; callers in `dispatch_sprints` and `Execute` bubble up. |
| 4 | `generate_code.go:118` — same root-of-write bypass | `args.OutputDir` clamped to `t.workDir` before any per-file `resolveUnderRoot` call. |
| 5 | `generate_code.go:205` — single-call mode missing fence-strip | Now applies `stripMarkdownFences` to `resp.Text()` before `parseFiles`, matching the sequential branch. |
| 6 | `write_enriched_sprint_example.md:972,998` — markdown-lint reference-link false positive | Wrapped both `body[0][...]` expressions in backticks. |

Build clean, all 17 packages green.

Sugi-chan: thanks again for `ee23412` — pr-feedback-3 + pr-feedback-4 + pr-feedback-5 should now cover everything from the consolidated review. Happy to take any further passes.

clintecker:
Pushed `d1a3f28` (pr-feedback-6) — handles Copilot's review of `186ce47`.

**Stale (already fixed in pr-feedback-5):** Copilot's review was submitted at 16:03 UTC, ~3 min before pr-feedback-5 landed at 16:06 UTC, so it was looking at the pre-symlink-aware snapshot. Both symlink comments (`completer.go:50` and `write_enriched_sprint.go:487`) are addressed by the `EvalSymlinks` work in pr-feedback-5.

**Real new finding:** `similarityRatio` mixed byte length (`len(a)`) with rune-based Levenshtein distance. For any non-ASCII content (em-dashes are everywhere in our sprint specs) the divisor overstates string length, producing artificially low ratios that could push genuine fuzzy matches below the 0.9 accept threshold. Fixed: divisor now `len([]rune(...))`. Regression test added.

CI was green on pr-feedback-5 (Quality Gates 2m4s, CodeRabbit pass) and should stay green here. All 17 packages still passing locally.

Issue #4, opened 2026-03-21
fix(pipeline): DippinValidated bypasses tracker validation without safety contract


harperreed:
## Description

`FromDippinIR` unconditionally sets `g.DippinValidated = true` (dippin_adapter.go:70), which causes `validate.go:78` to skip five validation categories: start/exit presence, edge endpoint integrity, exit outgoing edges, reachability, and cycle detection.

This trusts a pre-1.0 external library (`dippin-lang v0.1.0`) without defense-in-depth. If dippin-lang's validator has a bug, tracker silently accepts invalid graphs.

## Flagged By

5 of 6 reviewers (Go Purist, HN Commenter, Standards Zealot, Error Path Pessimist, API Surface Lawyer, Test Skeptic)

## Suggested Fix

Options (in order of preference):
1. **Always run tracker's own validation** — if checks are redundant with dippin, great (they're fast). If they catch something dippin missed, that's defense in depth.
2. **Rename to `PreValidated`** and make it format-agnostic on Graph, removing the dippin-specific naming leak.
3. **Encode which rules were covered** (bitmask or rule list) instead of a blanket boolean.

Additionally, `DippinValidated` is an exported field on `Graph` — any consumer can set it to `true` and silently bypass validation. This is a contract violation.

## Related

The `ensureStartExitNodes` function also silently mutates node shapes/handlers without warning, which is masked by the validation skip.

PR #201, opened 2026-05-05
fix(writes): self-healing extraction cascade for declared writes

harperreed:
## Summary

- When an LLM responds with prose instead of JSON, `applyDeclaredWrites` now tries a 3-layer extraction cascade before giving up: fenced ```json blocks, outermost `{…}` substring, then single-key fallback to raw response
- Single-key writes that can't find any JSON fall back to using the raw response as the value with a `writes_warning` (not `writes_error`), keeping the pipeline alive
- Multi-key writes with no parsable JSON still hard-fail since there's no way to distribute prose across multiple keys

**Root cause:** The NIFB spec-to-sprints run (`harper-nifb`, run `c78eb408d6d4`) failed at `analyze_spec` because the agent wrote `.ai/spec_analysis.md` successfully but responded in prose ("Done — ...") instead of JSON. The runner's `json.Unmarshal` hard-failed on the first character and killed the node.

## Test plan

- [x] 8 new tests for `ExtractJSONFromText` (fenced blocks, bare fences, outermost braces, nested braces, pure prose, invalid braces, fence priority, empty string)
- [x] 5 new tests for `applyDeclaredWrites` healing paths (fenced extraction, brace extraction, single-key fallback, multi-key still fails, direct JSON still works)
- [x] Updated existing tool handler test to verify single-key fallback + added new multi-key failure test
- [x] Full `go test ./...` passes (all 17 packages)
- [x] Pre-commit hooks pass

## Summary by CodeRabbit

* **New Features**
  * Improved JSON extraction from tool/LLM output: fenced code blocks, balanced top-level {…} spans, and NDJSON-first-object behavior.

* **Improvements**
  * Single-key declared writes can fall back to a capped raw response with a warning; multi-key missing/invalid JSON yields errors.
  * Non-object JSON (arrays/scalars) rejected; fenced blocks take precedence.
  * Validation rejects declared write keys that collide with reserved tool-command keys.

* **Documentation**
  * Updated docs and changelog describing cascade, failure modes, and fallback limits.

* **Tests**
  * Expanded coverage for extraction, fallback, validation, dedupe, and regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Self-healing JSON extraction cascade for declared writes** (PR #201). When an LLM responds with prose instead of valid JSON for a node with `writes:`, the runner now attempts: (1) direct JSON parse; (2) extraction of any ```...``` fenced block whose content parses as a JSON object — iterating fences so a `text`/`bash` preamble doesn't block discovery of a later `json` fence; (3) balanced-brace scan for the first top-level `{…}` span that parses as an object (handles prose with stray brace pairs around real JSON without picking the wrong span); (4) single-key fallback to the raw response with a `writes_warning` (not `writes_error`) so the pipeline survives. Multi-key writes still hard-fail since prose can't be distributed. The fallback is gated on "no extractable JSON found" — a model that returned valid JSON missing the declared key gets a hard contract failure with a specific error, not a silent fallback. Fallback values are capped at 8 KiB to keep large tool stdout out of `status.json` / `activity.jsonl` / checkpoints.
+
+### Changed
+
+- **`writes:` declarations are rejected when they collide with the `tool_command` safe-key allowlist** (`outcome`, `preferred_label`, `human_response`, `interview_answers`) (PR #201). Previously a workflow author could declare `writes: outcome` on an agent or tool node and funnel LLM-controlled content into a reserved name, bypassing the sanitization that exists exactly to keep LLM output out of shell input. The new `pipeline.IsToolCommandSafeCtxKey` accessor is checked before any value is written; collision sets `writes_error` and fails the node. No existing pipelines used these collisions.
+
 ## [0.24.2] - 2026-05-03
 
 ### Fixed
--- a/docs/architecture/context-flow.md
+++ b/docs/architecture/context-flow.md
@@ -107,11 +107,17 @@ tool ExtractMeta
 
 Runtime contract:
 
-- If `writes:` is declared, the node output must be valid top-level JSON object for extraction.
-- Every declared key must be present; missing keys hard-fail the node.
+- If `writes:` is declared, the runner runs an extraction cascade against the node's output:
+    1. **Direct JSON parse** — the response IS a top-level JSON object. The fast path.
+    2. **Fenced-block extraction** — the response contains one or more ```...``` blocks; the runner walks every fence pair and accepts the first whose content parses as a JSON object. A non-JSON preamble fence (`text`, `bash`, etc.) does not block discovery of a later `json` fence.
+    3. **Balanced-brace scan** — no fences, but a top-level `{…}` span exists in mixed prose. The scanner finds the first balanced span (skipping `{` inside `[…]` arrays and inside JSON string values) that parses as a JSON object.
+    4. **Single-key prose fallback** — only when no JSON object was found by any prior step AND the node declares exactly one writes key, the raw response is stored under that key with a `writes_warning` set in the context. The fallback value is capped at 8 KiB. Multi-key writes do not fall back — prose can't be distributed across multiple keys.
+- Every declared key must be present in the extracted JSON. A model that returns valid JSON missing the declared key hard-fails the node with a specific `writes_error` ("contained extractable JSON but failed the writes contract") — the prose fallback does NOT fire in this case, since the model intentionally produced JSON.
+- `writes:` keys cannot collide with the `tool_command` safe-key allowlist (`outcome`, `preferred_label`, `human_response`, `interview_answers`). The runner rejects such declarations at runtime to prevent LLM output from landing in reserved names that bypass shell-input sanitization.
 - String fields are stored as plain strings; non-strings are stored as compact JSON text.
-- Extra produced fields are allowed and surfaced as warnings.
-- Backward-compatible built-ins (`last_response`, `tool_stdout`, `interview_answers`, etc.) are still written.
+- Extra produced fields (in the JSON but not declared) are allowed and surfaced via `writes_warning`.
+- Two reserved context keys carry the cascade signal: `writes_error` (hard failure — the node returned `OutcomeFail`) and `writes_warning` (cascade healed but with caveats — extras present, fallback used). Both are plain context keys; downstream nodes can branch on them via `${ctx.writes_error}` / `${ctx.writes_warning}`.
+- Backward-compatible built-ins (`last_response`, `tool_stdout`, `interview_answers`, etc.) are still written regardless of the cascade outcome.
 
 For human interview mode, `writes:` extraction uses the collected interview answers object (question IDs and normalized question-text keys mapped to answers).
 For other human modes (`freeform`, `choice`, `yes_no`), `writes:` extraction is not applied automatically; rely on built-in human response keys unless handler behavior is extended.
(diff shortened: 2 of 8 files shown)


clintecker:
Pushed `840c879` — addresses the consolidated review (CodeRabbit + PAL gpt-5.2 + 5-agent squad: correctness, security, test-coverage, API contract, Go idioms).

## Critical bugs fixed

| # | Bug | Sources | Fix |
|---|-----|---------|-----|
| 1 | `extractFencedJSON` early-exits on first fence (a ```text preamble blocks discovery of a later ```json fence) | CodeRabbit, PAL, Test-Coverage, Correctness | Iterate fence pairs via `strings.Cut`; return first valid-JSON-object content |
| 2 | Single-key fallback fires when `ExtractJSONFromText` returned valid JSON missing the declared key (silently ships prose under a name it doesn't fit) | CodeRabbit, PAL, Test-Coverage | New `foundExtractableJSON` flag gates the fallback to "no JSON found at all"; the missing-key path now hard-fails with a specific message |
| 3 | Fallback value uncapped — full tool stdout (up to 10MB) lands in context key, propagates to status.json, activity.jsonl, checkpoints, downstream prompts | Correctness | `maxFallbackValueBytes = 8192` cap with truncation marker |
| 4 | `writes:` collision with the `tool_command` safe-key allowlist (`outcome`, `preferred_label`, `human_response`, `interview_answers`) bypasses the shell-input sanitization gate | Correctness | New `pipeline.IsToolCommandSafeCtxKey` accessor; `applyDeclaredWrites` rejects collisions before any value is written |

## Other improvements

- `extractOutermostJSON` → `extractBracedJSON` (balanced-brace scanner). Skips `{` inside `[…]` (so `[{"a":1}]` isn't silently decomposed) and inside JSON string values (so paths like `"/tmp/{name}.txt"` don't break depth counting). The old first-`{`/last-`}` shortcut returned `""` on prose like `wrote {file1} and {file2}, then produced {"x":1}`.
- `isJSONObject` helper centralizes the "JSON object, not array/scalar/null" check.
- Doc comments now specify "JSON *object*" rather than just "JSON".
- `docs/architecture/context-flow.md` section 110-114 was stale ("missing keys hard-fail"); now describes the four-step cascade, the gating rule preventing fallback after partial extraction, the safe-key collision rejection, and the `writes_warning` / `writes_error` semantics.
- `CHANGELOG.md` updated under `[Unreleased]` per CLAUDE.md mandate.

## Regression tests added

`TestExtractJSONFromText_MultipleFences_FirstNotJSON`, `_StrayBackticksBeforeFence`, `_BracedSpanWithStrayPairs`, `_BracesInsideStringValue`, `_NDJSONReturnsFirstObject`, `_RejectsArrayAndScalar`. Plus `TestApplyDeclaredWrites_ExtractedJSONMissingKeyHardFails`, `_RejectsSafeKeyCollision` (sub-tests for all four reserved keys), `_FallbackValueIsCapped`.

## Findings deferred (squad agreed worth a follow-up but not blocking)

- `writes_warning` has no TUI / `tracker diagnose` / log surface (PAL + API contract).
- Prose fallback delivers non-JSON to consumers expecting JSON shape (e.g. `${ctx.x | jq}`). Documented but not hardened — would require a `writes_strict` mode or per-key shape declaration.
- The `tracker_doctor.go` `maxTok 16→1` drive-by remains in this branch (API contract flagged as unrelated; left at author's call).

## Security note

The security reviewer found **no critical or high** issues. The pre-existing `pipeline/expand.go` safe-key allowlist already blocks LLM-origin keys from `tool_command` interpolation; the fallback expanded an existing surface but didn't create a new sink. The one residual concern (workflow author colliding with the allowlist via `writes:`) is now closed by fix #4 above.

All 17 packages green under `go test ./... -short`. gofmt clean.

clintecker:
Pushed `f0f692c` — restores `maxTok := 16` in `tracker_doctor.go` to match current main.

Correction to my earlier message: I called the `maxTok 16→1` line a "drive-by." That was wrong. What actually happened:

- `maxTok := 1` predates this branch and was a real bug.
- @mdagost's PR #199 (merged earlier today) fixed it back to `16`. The commit message:
  > OpenAI's Responses API requires `max_output_tokens >= 16`, so `1` returns HTTP 400 with `Invalid 'max_output_tokens': integer below minimum` and the `tracker doctor` command breaks.
- This branch was forked from main *before* #199 merged, so it still carried the broken `1`. Without `f0f692c`, merging #201 would re-introduce the bug.

Thanks for the catch — apologies for the misread.

clintecker:
Rebased onto current main (force-pushed with lease). The branch was forked before #197/#198/#199/#200 merged today; rebasing pulls those into the base and drops `f0f692c` (the manual maxTok=16 re-application) since git correctly detected #199 already has that change.

Branch is now `2fa28aa fix(writes): address consolidated review of PR #201` on top of `30e00cf` (latest main with #198 merged).

Two commits ahead of main now, instead of three. CI will re-run.

clintecker:
Pushed `32e932f` — PAL re-review (gpt-5.2) of the post-rebase branch surfaced three minor improvements.

**Applied:**

1. **Anchored `fencedBlockRE`** to `(?:^|\n)[ \t]*` so the opening fence must start at the beginning of input or follow a newline (with optional leading whitespace). Defense-in-depth: a stray `X\`\`\`json\n...` mid-word can no longer in principle match, even if the rest of the language-tag-then-newline shape happens to align.
2. **Documented the regex limitation** in a code comment: a literal `\`\`\`` inside a JSON string value truncates the body since the regex isn't JSON-aware. The cascade falls through to `extractBracedJSON` (which IS string-aware via `matchBalancedBrace`) and recovers the object. The behavior was correct end-to-end but undocumented.
3. **Three tests pinning previously-implicit behavior:**
   - `TestExtractFencedJSON_TwoValidFences_FirstWins` — locks in the "first valid fence wins" semantic.
   - `TestExtractFencedJSON_TripleBacktickInsideStringFallsThroughToBraces` — proves the regex limitation in half 1 (extractFencedJSON returns `""`) AND the cascade rescue in half 2 (ExtractJSONFromText recovers via extractBracedJSON).
   - `TestExtractJSONFromText_TruncatedFenceFallsThrough` — deterministic behavior on an unclosed fence.

**Skipped (PAL flagged, not worth this PR):**
- Tightening the safe-key error wording (current message is clear).
- Lint-time check for safe-key collision (real follow-up; separate PR).
- More commentary on the 8 KiB cap being non-configurable.

Bonus: while tracing edge cases, I caught a real bug in my own previous fix (`cec516f`) — `TestExtractJSONFromText_StrayBackticksBeforeFence` was passing only because `extractBracedJSON` rescued it; the fenced extractor itself was still broken on stray backticks. The new white-box test in `cec516f` (and the regex fix) closed that hole. Without PAL's re-review nudge I would have shipped a maintenance hazard.

All 17 packages green, gofmt clean. CI re-running.

clintecker:
Pushed `9aa9395` — addresses both findings from the follow-up pass.

**Finding 1 (inline) — applied with a clarification.** Reserved `writes_error` and `writes_warning` as authoring-error keys.

The original collision check covered the tool_command safe-key allowlist only (`outcome`, `preferred_label`, `human_response`, `interview_answers`). That's the security set — the ctx.* keys allowed in tool_command shell expansion. `writes_error` and `writes_warning` aren't on that allowlist, so a writes-key collision with them doesn't bypass shell sanitization (the original "spoof" framing was technically inaccurate). But they ARE runtime-owned observability signals — `applyDeclaredWrites` sets them so `tracker diagnose` and downstream `when ctx.writes_error != ""` edges can branch on them. Letting a workflow declare `writes: writes_error` would let the LLM fake a failure signal (or hide a real warning). Different concern from the security one, but a real integrity issue.

Added an `isReservedWritesSignalKey` helper; the collision check now spans both sets with a single unified error message naming both categories. Test table extended:

```
TestApplyDeclaredWrites_RejectsReservedKeyCollision
  ├── outcome           (tool_command safe-key)
  ├── preferred_label   (tool_command safe-key)
  ├── human_response    (tool_command safe-key)
  ├── interview_answers (tool_command safe-key)
  ├── writes_error      (writes signal)
  └── writes_warning    (writes signal)
```

(Renamed from `_RejectsSafeKeyCollision` to reflect the broader scope.)

**Finding 2 (nitpick) — applied.** Added `tool_stdout` assertions to both `TestToolHandlerDeclaredWritesSingleKeyFallsBackToRaw` and `TestToolHandlerDeclaredWritesMultiKeyInvalidJSONFails`. Verified the contract: `ContextKeyToolStdout` is set by `tool.go:335` *before* `applyDeclaredWrites` runs, and the writes processing only adds keys — so stdout is preserved today. But that contract was invisible to the test suite; a future refactor could lose it without any test failing. Now pinned.

Branch is at `9aa9395`, five commits ahead of main. CI re-running.

Issue #233, opened 2026-05-15
build_product: 8 workflow gaps that let CI-red, wrong-shape API calls, and Phase-6 scope-creep through the cross-review/final-check gates


clintecker:
# `build_product`: 8 concrete workflow gaps that let CI-red, wrong-shape API calls, and Phase-6 scope-creep through the cross-review/final-check gates

## What this issue is about

The `examples/build_product.dip` workflow ran end-to-end on a real Phase 1 spec (the `code-goblin` daemon) and reported the work as Done. A subsequent independent five-lane audit found:

- **CI was red on the "done" commit** (`make lint` exits non-zero with 13+ violations).
- **The first real OpenAI call would 400** (request body uses Chat Completions `response_format` shape instead of Responses API `text.format`).
- **A Phase-6 feature shipped in Phase 1** (comment trailer `fingerprints` field populated when the spec explicitly says empty).
- **Retries off-by-one, `gh pr list` invocation wrong on two axes, several interface methods defined-and-tested but never called from production**, plus ~30 smaller deviations.
- **Many of these bugs are actively asserted by green tests.** Same-agent-wrote-impl-and-test failure mode; the assertion pins the wrong behavior.

The audit caught these. The workflow didn't. **This issue is about the workflow, not about the symptoms** — the symptoms are evidence and regression test cases. The actual code fixes belong in a separate issue against `2389-research/code-goblin` (and we can file one once the workflow fix lands).

The root question for tracker maintainers is: **what does `build_product` need to be changed to do, so that its next run on a comparable spec catches these classes of error before declaring Done?**

I'll number the gaps and propose specific changes for each. The audit findings are preserved in a collapsible section at the bottom — each is tagged with which gap(s) would have caught it.

---

## How `build_product` failed, step by step

I read every per-step prompt from this run (`.tracker/runs/423c926417d0/<step>/prompt.md` and `response.md` for `Decompose`, `Implement`, `TestMilestone`, `VerifyMilestone`, `ReviewClaude`, `ReviewCodex`, `ReviewGemini`, `SynthesizeReviews`, `FinalBuild`, `FinalSpecCheck`) plus the workflow definition at `examples/build_product.dip`. Citations below point to the workflow file unless otherwise noted.

### Gap 1. `FinalBuild` (and `TestMilestone`) never run the project's own CI command

**Evidence — `examples/build_product.dip`:**

```
tool FinalBuild
  label: "Final Build & Test"
  timeout: 300s
  command:
    set -eu
[...]

PR #246, opened 2026-05-19
feat(build_product): close Gaps 1+3+6 from #233 audit

clintecker:
## Summary

Address the three lowest-cost, highest-leverage gaps from #233 — the
audit that found 38 issues `build_product` declared "Done" on. The
issue's recommended ordering names Gaps 1, 3, 6 as the cheap trio that
catches most of the symptoms (red CI, wrong-shape API calls, off-by-one
retries, Phase-6 features in Phase 1).

The remaining gaps (2 reviewer rubric, 4 VerifyMilestone reads SPEC.md,
5 engine `auto_status` audit, 7 interface reachability, 8 TestQuality
step) are out of scope here; the next session will pick them up.

## Files touched

- `examples/build_product.dip` — workflow edits (Gaps 1, 3, 6).
- `workflows/build_product.dip` — embedded copy kept in sync via
  `make sync-workflows` (CI enforces parity).
- `CHANGELOG.md` — Unreleased / Changed entry.

### Gap 1 — project CI is now part of the gate
`TestMilestone` and `FinalBuild` previously ran only the language-stack
default (`go build && go test`, `npm test`, etc.) and silently passed
even when the project's own `make lint` was red. Both nodes now probe
for a Makefile and run the first defined target out of `ci` / `check`
/ `lint`, gating on its exit.

**Probe design** (revised after PR review — Codex P1, CodeRabbit Major,
Copilot): the first draft used `make -n <target>` for existence
detection. That's unsafe — GNU Make returns 0 when the target name
matches a file/directory on disk (so `ci/` and `lint/` directories
would false-positive), and a missing `make` binary or a Makefile
parse error would collapse silently to "target absent." The current
probe greps the Makefile directly for top-level target declarations
(`^[ ]*<TARGET>[ ]*:` after stripping comments; recipe lines are
tab-prefixed and don't match). Both nodes also verify `command -v
make` before invoking it — when a Makefile is present but make
isn't installed, the node fails loud rather than skipping the gate.
Trade-off: misses targets defined only in included files; acceptable
since CI/lint/check are by convention in the root Makefile.

When no Makefile or no matching target exists, the node emits a
visible `INFO:` line so operators reading `tracker diagnose` see the
gap rather than assume CI passed. `FinalBuild`'s timeout bumped
300s → 600s for the extra lint pass.

### Gap 3 — `Implement` anchors on spec literals and test quality
Added three rule blocks to the prompt:
- **Spec literals are contracts**: for every literal in the spec
  section (command strings, JSON keys, header names, integer
  constants, log keys), grep the implementation byte-for-byte before
  committing. Silent paraphrase (`--head <branch>` vs spec's
  `--head <owner>:<branch>`) is the most common way milestones ship
  wrong.
- **Tests verify the contract, not your code**: every assertion must
  still pass for the right reason if the production code is deleted
  and rewritten spec-conformantly. The canonical failure mode is
  asserting `attempts == 2` when the spec says "max 2 retries"
  (= 3 attempts).
- **Snapshot tests must be hand-verified**: the FIRST version of every
  golden file must be anchored to the spec, not regenerated from
  current output.

### Gap 6 — `Decompose` produces a DO NOT list, `Implement` reads it
New milestone field `**DO NOT implement**:` names Phase 2+ features
the spec defers but whose supporting types/functions live in this
milestone's file list. `Implement` reads
`.ai/milestones/current.md`'s DO NOT lines before writing in any file
and leaves those affordances inert. Closes the entire
"wrote-future-phase-with-green-tests" failure class — e.g., the
`fingerprints` field the audit found populated in Phase 1 despite the
spec deferring it to Phase 6 would now appear as a DO NOT line.

## Test plan
- [x] `dippin doctor examples/build_product.dip` — **A / 100/100**, no
      new lint warnings, 25 nodes all reachable
- [x] `tracker validate examples/build_product.dip` — clean
- [x] `tracker simulate examples/build_product.dip` — nodes + edges
      unchanged
- [x] `go build ./...`
- [x] `go test ./... -short` — all packages pass
- [x] `make check-workflows` — embedded `workflows/build_product.dip`
      stays byte-identical to `examples/build_product.dip`
- [x] CI probe regression cases verified end-to-end:
  - empty Makefile + `ci/` directory: probe correctly returns no
    match (was the Codex P1 false-positive)
  - tracker's own Makefile: probe correctly picks `make ci`
  - commented `# ci:` line: no false positive
  - tab-prefixed recipe line `\tci:`: no false positive

Refs #233

## Summary by CodeRabbit

* **Changed**
  * Project CI gate discovery now auto-detects and runs top-level Make targets (ci → check → lint) when present; errors if a Makefile exists but make is unavailable and logs when no gate target is found.
  * Per-milestone "DO NOT implement" lists added; implementers must read them and keep deferred Phase‑2+ affordances inert while allowing type/signature placeholders.
  * Spec-literal anchoring enforced (byte-for-byte matching) and test guidance tightened to validate spec contracts; initial golden/snapshot results require hand verification.
  * Final build timeout increased from 5 to 10 minutes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/2389-research/tracker/pull/246?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


clintecker:
## Re: round-6 review (false positive — regex is correct)

Copilot's 3 round-6 comments claim that `substr($0, cp) ~ /^:+=/` doesn't match `::=` / `:::=`. **This is not correct** — `:+` is a greedy quantifier that matches 1 or more colons before requiring `=`. I verified directly:

```
$ for input in ":=" "::=" ":::=" ":" ": dep" "::"; do
    result=$(echo "$input" | awk '{ if ($0 ~ /^:+=/) print "MATCH"; else print "NOMATCH" }')
    printf "  %-12s → %s\n" "$input" "$result"
  done
  :=           → MATCH
  ::=          → MATCH
  :::=         → MATCH
  :            → NOMATCH
  : dep        → NOMATCH
  ::           → NOMATCH
```

End-to-end probe test on real Makefile inputs (each `<assignment>` is a one-line Makefile, probe target = `ci`):

```
  ci := value → NOMATCH (correct)
  ci ::= value → NOMATCH (correct)
  ci :::= value → NOMATCH (correct)
  ci:=value → NOMATCH (correct)
  ci::=value → NOMATCH (correct)
  ci:::=value → NOMATCH (correct)
```

The probe correctly skips all three POSIX/GNU assignment forms; no fix needed.

The inline comment says "first ':' part of := / ::= / :::=" — that's accurate. The CHANGELOG says "skips variable assignments by checking whether the first `:` is part of `:=` / `::=` / `:::=` via `^:+=`" — also accurate.

Not making a change for this round; happy to revisit if there's a concrete Makefile that demonstrates the claimed regression.

Issue #75, opened 2026-04-15
feat: dippin lint should recognize ctx.node.<id>.* scoped reads


clintecker:
With per-node context scoping (PR #73, issue #32), tracker now supports reads like \`\${ctx.node.MyAgent.last_response}\` and declarations like \`reads: [node.MyAgent.last_response]\` in .dip files.

However, dippin-lang's lint step treats these as undefined context keys, producing false-positive warnings even though the keys are valid in tracker.

The fix: update dippin-lang's lint to recognize keys matching the pattern \`node.<id>.*\` as implicitly available (or add a pattern-based reserved-key check), since tracker provides these automatically after each node execution.

This is a cross-repo concern but blocks workflows from passing dippin doctor cleanly when they use scoped context reads.

clintecker:
Filed upstream: https://github.com/2389-research/dippin-lang/issues/20

clintecker:
**Status update: upstream merged, awaiting release.**

dippin-lang [PR #23](https://github.com/2389-research/dippin-lang/pull/23) taught the lint to recognize `ctx.node.<id>.*` as valid scoped context reads (merged 2026-04-17 20:13 UTC). The fix is on dippin-lang `main` but landed after v0.20.0 shipped.

**To consume:** bump `go.mod` to the next dippin-lang release (v0.21.0+). No tracker code changes needed — the lint behavior is internal to dippin-lang.

Close this issue once the dep bump lands and `dippin doctor` on a workflow using `${ctx.node.<id>.*}` references returns no false-positive warnings.

clintecker:
Also tracked upstream at [dippin-lang#25](https://github.com/2389-research/dippin-lang/issues/25) — open issue requesting the v0.21.0 release cut.

PR #273, opened 2026-06-01
feat: validation_overridden terminal status (Gap 5.2, #271)

clintecker:
## Summary

- Adds a new terminal `EngineResult.Status` value **`validation_overridden`**, triggered by per-edge `override: true` markings on `wait.human` gate edges. Audit-only signal — `ctx.outcome` routing inside a run is unchanged.
- Threads the override through the full stack: engine flip-point (with sticky list + crash-resume durability), subgraph/parallel/manager_loop propagation, library API (`tracker.Result.ValidationOverrides`, `AuditReport`/`RunSummary`/`DiagnoseReport` mirrors), CLI (audit header chain, list table, summary amber, `--fail-on-override` flag + exit code 2 + `TRACKER_FAIL_ON_OVERRIDE=1`), TUI live rendering (amber completion row with gate/label/actor).
- Bundles a long-standing fix: `tracker_audit.classifyStatus` previously collapsed `budget_exceeded` events to `"fail"` (D12 in the spec) — audit and list surfaces now surface `budget_exceeded` correctly. Audit recommendations no longer alphabetically sorted; override notes take priority order (D16).
- Adds `dippin doctor` lint rule **TRK102** that warns on unmarked override-shape edges using a 4-predicate heuristic (suppresses `ApprovePlan`-shape false positives via the upstream-fail check).
- Adds the `status_class` companion field (one of `succeeded`/`failed`) to `cmd/tracker-conformance` JSON for stable bucketing across future enum extensions (D15).
- Closes Gap 5.2 of #233. Tracking issue: #271.

## Design and plan

- **Spec:** `docs/superpowers/specs/2026-05-29-validation-overridden-design.md` (v2, 950 lines, 18 decisions D1-D18, brainstormed and twice-reviewed by 6+4-agent panels).
- **Plan:** `docs/superpowers/plans/2026-05-29-validation-overridden.md` + `.part2.md`. 32 tasks across 14 chunks. Executed via `superpowers:subagent-driven-development` with fresh implementer + spec reviewer + code quality reviewer per task.

## What's not in this PR (blocked on dippin-lang)

Four tasks (Chunks 4 + 13, Task 26) require a tagged `2389-research/dippin-lang` release that adds `ir.Edge.Override` so `.dip` files can parse `override: true`. They land in follow-ups:

[...]

PR #275, opened 2026-06-01
feat: writable_paths fs-jail enforcement (#272)

### LINKED ISSUE #272: feat: enforce writable_paths fs-jail (paired with dippin v0.35.0 / #75)
Runtime enforcement for the dippin **`writable_paths`** authoring primitive. dippin carries + lints the field; **tracker must enforce it**. This is a **joint release**: dippin's side is merged (dippin PR #83, issue #75) but is intentionally **unreleased** until this lands — shipping the authoring surface without enforcement is a silent fail-open on a safety field (the "lint-validated runtime-no-op safety field = worse than nothing" anti-pattern from dippin #41).

**Dippin spec (normative contract — read this):** `docs/superpowers/specs/2026-05-29-issue-75-writable-paths-design.md` in 2389-research/dippin-lang (merged on `main`). The § "Required tracker enforcement contract", § "Tracker-side tests", and § "Release coordination" sections are the source of truth.

## What the field is
`writable_paths` is a comma-separated glob list on agent nodes and parallel branches (e.g. `writable_paths: .ai/sprints/**, .ai/managers/recovery-journal.md`). It bounds where an agent's tools may write. It reaches tracker via the packed `.dipx` (re-parsed). Motivating sites: the 5 failure/recovery recorders in 2389-research/pipelines (PR #23/#25 — the `reasoning_effort: high` agent that wrote arbitrary Rust + ran `cargo` despite a prompt restriction).

## Enforcement requirements (normative)
1. **Write containment (native backend).** Every file mutation by ANY tool the agent can invoke — `Write`/`Edit`/`ApplyPatch` **and `Bash` and any process Bash spawns** — must be confined to paths matching `writable_paths`, resolved against the session root. Tool-argument validation alone is insufficient; Bash is a write primitive.
2. **Symlink-chain resolution at write time.** Reject a write whose resolved real path (full parent symlink chain, EvalSymlinks-equivalent) escapes the allowed set — including a symlink the agent forges at runtime inside an allowed dir. Close the TOCTOU (check on the resolved path at write time). (Mirrors dippin #67/#77.)
3. **Immutable anchor.** Globs resolve against a session root fixed at session creation. `working_dir` (a first-class AgentConfig field) and any `Params` key MUST NOT relocate the anchor.
4. **Fail-closed.** A `writable_paths` that is present-but-empty, malformed/uncompilable, or **unrecognized (version skew — an older tracker)** → **deny-all-writes or refuse-to-start**, never unbounded. (dippin already rejects present-but-empty at parse/validate, but tracker must still fail-closed on anything malformed/unknown it receives.)
5. **Bypass defense.** When `writable_paths` is set, do not honor a `Params` key or the `working_dir` field that widens/relocates the write surface. (Mirrors #41's Params bypass defense.)
6. **Backends.** Enforce on **native** only in v1. On `claude-code`/`acp` (where tracker doesn't control the child sandbox), **refuse session creation** when `writable_paths` is set — do NOT degrade to prompt-level `--allowedTools` (that's the prompt-discipline failure this primitive replaces).

### Out of scope (residual escapes — document, don't silently imply handled)
Network (`curl`/`cargo` fetch), reads/exfil-by-read, content-within-an-allowed-path (blast radius of e.g. `workspace/**`), hardlinks / inherited FDs / `/proc` re-entry. Name the residuals in the PR.

## Required tests (restore #41 red-team rigor)
- **Red-team, single-turn:** agent `writable_paths: workspace/**`; mock LLM emits, in ONE response, `write("../escape.rs", …)`, `bash("echo pwned > /etc/x")`, `bash("cargo build")`. Assert zero writes outside `workspace/**`, and that a **child process spawned by Bash** writing outside is also blocked.
- **Runtime symlink escape:** `bash("ln -s /etc workspace/link && echo x > workspace/link/y")` → denied.
- **working_dir relocation:** `writable_paths: workspace/**` + `working_dir: /` (or `..`) → still bounded to the original anchor.
- **Fail-closed:** present-but-empty / malformed glob / unrecognized-field → deny-all or refuse, never unbounded.
- **Bypass:** Params working-dir/permission override ignored.
- **Branch inherit:** agent `writable_paths: workspace/**` fanned out through a branch that OMITS the field → still bounded (never unbounded). (Effective = branch if non-empty else agent.)
- **Backend:** native enforces; claude-code/acp refuse to start.

## Release coordination (ordering matters)
1. This tracker PR merges + **tracker tag cut first**.
2. Then dippin cuts **v0.35.0**, pinning `requires tracker >= <this tag>` as a **safety** requirement in CHANGELOG/skill.md, and records the tracker SHA/tag in the dippin spec.
3. The tracker team bumps their dippin pin to v0.35.0 to adopt.

Until this ships and is tagged, dippin v0.35.0 must NOT be released.

_Filed from dippin #75 / PR #83 follow-through. Contract details and the full red-team table are in the dippin spec linked above._

clintecker:
## Summary

Implements tracker's runtime enforcement of dippin's `writable_paths` primitive (paired with dippin-lang #75): a Linux Landlock fs-jail bounding both Bash subprocess + descendants AND in-process Write/Edit/ApplyPatch to author-declared globs resolved against the session root. Closes #272.

**Two-tier semantic** (spec § D2):
- **In-process tools** enforce the *exact* globs via `openat2(RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS)` against a session-root fd — symlink chains rejected at the kernel syscall, no TOCTOU window.
- **Bash subprocess** is bounded at directory ancestors of each glob's static prefix via Linux Landlock ABI v3 (kernel 6.7+), inherited through `syscall.Exec` so every descendant (cargo, rustc, child shells) is sandboxed.

**Three refuse-to-start gates** at session setup:
- G1 — bad `working_dir` (escapes process cwd) or malformed/empty/absolute/parent-escape globs
- G2 — backend ∈ {claude-code, acp, unknown} (out-of-process; tracker cannot sandbox)
- G3 — Landlock unavailable (non-Linux or kernel < 6.7)

**Bypass defenses**:
- Adapter writes empty marker when typed IR field is empty but `Params` carries `writable_paths` — blocks Params-spill bypass.
- `WritablePathsSet` companion field preserves the fail-CLOSED signal across IR → adapter → SessionConfig boundaries.
- `Pdeathsig=SIGKILL` on every spawned subprocess: if the parent dies for any reason, the kernel atomic-reaps the children — closes the orphan-accumulation hole that surfaced during dev (132 live `__jail-exec` orphans, load avg 74 on a 4-core box).

**Internal subcommand** `tracker __jail-exec` — dispatched before flag parsing in `cmd/tracker/main.go`. Re-execs via `/proc/self/exe` to apply Landlock then `syscall.Exec` into the agent command. Operators MUST NOT invoke directly; documented for transparency.

## Implementation overview

| Layer | Files | What |
|---|---|---|
| **IR plumbing** | `pipeline/node_config.go`, `pipeline/dippin_adapter.go`, `agent/config.go` | Typed `WritablePaths` + `WritablePathsSet` carry from dippin IR → AgentNodeConfig → SessionConfig with typed-wins precedence + Params-bypass defense |
| **Env seams** | `agent/exec/local.go` | Optional `CommandWrapper` + `WriteOpener` function fields on LocalEnvironment — zero behavior change when nil |
| **Jail (Linux)** | `agent/exec/jail_linux.go`, `agent/exec/jail.go`, `agent/exec/jail_errors.go` | `ProbeLandlock` (ABI v3 floor) · `WrapBashCmd` (argv rewriter) · `RunJailExec` (self-re-exec target with strict Landlock + RODirs("/") for sh access) · `OpenForWrite` (openat2 + glob matching) · `ValidateWritablePaths` (cross-platform path math) |
| **Jail (non-Linux)** | `agent/exec/jail_other.go` | Passthrough stubs that fail-closed; refuse-to-start when `writable_paths` declared |
| **Handler wiring** | `pipeline/handlers/codergen_jail.go`, `pipeline/handlers/codergen.go`, `pipeline/handlers/backend_native.go` | `configureJail` 3-gate refuse-to-start; per-session fresh `*LocalEnvironment` (never mutates shared `b.env` across concurrent branches) |
| **Dispatch** | `cmd/tracker/main.go` | `__jail-exec` short-circuit before `parseFlags` |
| **Hardening** | `agent/exec/parent_death_linux.go` | `Pdeathsig=SIGKILL` on every spawned cmd — kernel-level orphan reaper |

## Test plan

- [x] `go build ./...` clean on Linux + `GOOS=darwin GOARCH=amd64` cross-build
- [x] `go test ./agent/exec/... -short` — all jail tests pass on kernel 6.17 / Landlock ABI v8
- [x] `go test ./pipeline/handlers -short -race` — race detector clean
- [x] Red-team kernel-level: `TestRunJailExec_DeniesOutsideWrite`, `TestRunJailExec_DeniesSymlinkEscape`, `TestRunJailExec_AllowsInsideWrite`
- [x] In-process exact-glob enforcement: `TestInProcessWriteEnforcesExactGlob`, `TestOpenForWrite_RejectsParentEscape`, `TestOpenForWrite_RejectsSymlinkEscape`, `TestOpenForWrite_RejectsAbsolutePath`
- [x] Handler-layer E2E: `TestWritablePathsEnforcement` (3-row), `TestBranchEnforcesResolvedPaths`
- [x] Refuse-to-start gates: `TestConfigureJail_RefusesOn{ClaudeCode,Acp,UnknownBackend,InvalidPaths,NoLandlock}`, `TestWorkingDirRelocationRefused`, `TestWritablePathsFailClosed`
- [x] Bypass defense: `TestExtractAgentAttrs_WritablePaths/Params writable_paths is blocked when typed empty`
- [x] Parallel-branch race vector: `TestParallelBranchSymlinkRace` (bounded 100 iterations)
- [x] Orphan-reaper defense: `TestApplyParentDeathSig_LocalEnvironment` confirms `Pdeathsig=SIGKILL` on both Exec paths

## Known intentional state

- `TestPinnedDippinVersionMatchesGoMod` **fails** on this branch. The branch pins dippin-lang to `@latest` (pseudo-version) during cross-repo joint-release dev per the user's explicit instruction. The release PR will bump dippin-lang to the tagged `v0.35.0` and bump `tracker_doctor.go`'s `PinnedDippinVersion` in lockstep, restoring green.
- All other 17 packages pass `go test ./... -short` on a Linux 6.7+ host with Landlock support.

## Residual escape classes (documented in CHANGELOG)

Not bounded by `writable_paths`:
- Network egress (`curl`, `cargo` crate fetches)
- Reads / exfil-by-read
- Content within an allowed path (agent can still poison `workspace/Cargo.toml` if `writable_paths: workspace/**`)

Narrow globs are the strongest posture.

## Documents

- Design spec: `docs/superpowers/specs/2026-06-01-issue-272-writable-paths-enforcement-design.md`
- Implementation plan: `docs/superpowers/plans/2026-06-01-issue-272-writable-paths-enforcement.md`
- Architecture note: `CLAUDE.md` § `tracker __jail-exec` internal subcommand
- User-facing: `CHANGELOG.md` § [Unreleased]

---
<a href="https://app.blacksmith.sh/2389-research/codesmith/tracker/pr/275"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782939040&installation_id=131176874&pr_number=275&repository=2389-research%2Ftracker&return_to=https%3A%2F%2Fgithub.com%2F2389-research%2Ftracker%2Fpull%2F275&signature=051aa7cf9018923a7221c2d3b9c2fa78352a22e88d0c3dcd9767eb38e7ba762e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Runtime enforcement for writable_paths: file writes (including shell subprocesses) are restricted to declared globs; enforcement uses Linux Landlock (ABI v3, kernel 6.7+). Sessions fail-closed when unsupported or misconfigured.

* **Documentation**
  * Added design, plan, and changelog docs describing behavior, limits, and operator guidance.

* **Tests**
  * Added extensive unit, integration, and Linux end-to-end tests for validation, Landlock probing, and glob semantics.

* **Chores**
  * Updated platform/security dependencies and bumped the pinned integration version.


clintecker:
Comprehensive review-fix pass pushed (`d19897a`). 14 files, +314 / −70.

## Addressed

### Security (codex P1)
- **ApplyPatch delete bypass** (local.go:108): ApplyPatch's `delete` and move-cleanup paths called `os.Remove` directly, bypassing the jail. New `ExecutionEnvironment.RemoveFile` method + `env.Remover` optional hook + `configureJail` wires it with the same exact-glob check as `WriteOpener`. Destructive operations now bounded alongside writes.

### Correctness — jail layer
- **Recursive glob** (codergen_jail.go:129): `matchOneGlob` now handles `prefix/**/suffix`, `**/suffix`, and trailing-`**`. Previously only bare `**` and `/**` had recursive semantics, so file-targeting recursive globs fell through to non-recursive matching.
- **`..foo` false positive** (codergen_jail.go:82): replaced `HasPrefix(relPath, "..")` with `== ".." OR HasPrefix(relPath, "../")`. Valid filenames like `..cache` no longer rejected.
- **MkdirAll before glob check** (local.go:108, codex P2): mkdir now happens INSIDE the WriteOpener closure, AFTER the glob check. Rejected writes no longer leave empty intermediate directories outside `writable_paths`.
- **Anchor drift on node-level `working_dir`** (backend_native.go:57): jailed env now rooted at `sessionCfg.WorkingDir`, not the shared backend env's workDir. Prevents `cmd.Dir` / relative tool paths from drifting outside the anchor `configureJail` validates against.
- **`codergen` alias rejected by jail** (codex P2 codergen.go:634): `buildConfig` normalizes `cfg.Backend == "codergen"` → `"native"` before threading into SessionConfig. The G2 gate now accepts the documented alias.
- **Malformed character classes slip through** (jail.go:91): `validateGlobEntry` now probes with `path.Match` so patterns like `foo[` fail at config validation instead of becoming opaque runtime denials.

### Correctness — Pdeathsig (coderabbitai parent_death_linux.go:43)
- Real Go-runtime gotcha: `PR_SET_PDEATHSIG` is per-OS-thread, not per-process. If the Go runtime migrates the calling goroutine off its original thread and the thread is later retired (GC stop-the-world, thread-pool reduction), the kernel sends a premature SIGKILL to the child.
- Fix: new `pinCallingThreadForParentDeath()` helper does `runtime.LockOSThread()` / `UnlockOSThread()`. Applied at `ExecCommand`, `runUnlimited`, `runLimited` — the thread stays alive while the child runs.

### Lint
- **ST1005** (jail_errors.go:11): lowercase `landlock ABI v3...`.

### Tests
- **TestParallelBranchSymlinkRace rewrite** (e2e:310, e2e:311): branch B now writes through `branchA/share/payload-N.txt` — the SAME path branch A is mutating — so the TOCTOU is actually exercised. Vacuity guard requires ≥1 successful in-jail write so the test can't pass on a broken impl that rejects everything.
- **TestWritablePathsEnforcement** (e2e:97): startup errors on `env.ExecCommand` now fail the test. Negative rows can't pass on a missing-`sh` / jail-dispatch failure.
- **TestRunJailExec_Denies* positive side-effect** (jail_linux_test.go:100, 179-198): both negative-path tests now assert an allowed write OR symlink creation succeeded BEFORE checking the denial. Vacuous-pass on `sh` / re-exec startup failure is closed.

### Docs
- Plan doc `configureJail` signature: `(cfg, attrs, env)` → `(cfg, env, processCwd)` to match the implementation.

## Explicitly deferred

| Item | Reviewer | Why |
|---|---|---|
| go.mod pseudo-version pin | coderabbitai (🔴) | Documented in the PR description as the intentional cross-repo dev state; the release PR bumps both `go.mod` and `PinnedDippinVersion` in lockstep when dippin tags v0.35.0 |
| MD040 fence-language nits in plan doc | coderabbitai (🟡) | Low signal, docs-only |
| Shallow-copy hedge in engine_run.go | coderabbitai (#273) | Already deferred there; same rationale |

## Verification

- `go build ./...` clean on Linux + `GOOS=darwin GOARCH=amd64`
- `go vet ./...` clean
- `gofmt -l .` empty
- `go test ./... -short` — all 18 packages PASS
- `go test ./pipeline/handlers -run TestParallelBranchSymlinkRace -race` clean

clintecker:
Round 3 review pass pushed (`9950f13`, rebased + force-pushed). Three more Copilot items addressed; PR #273 was merged to main during this cycle so the branch is now rebased onto its post-merge state.

## Addressed

### Symlink-safe destructive operations (Copilot codergen_jail.go:92, :103)

**Real correctness gap.** `os.MkdirAll` inside the `WriteOpener` closure followed symlinks in intermediate components — an agent who pre-created `workspace/` as a symlink to an external directory could trick the jail into creating directories outside the anchor even though the final `openat2` would refuse the leaf file. `os.Remove` in `env.Remover` had the analogous hole on the delete path.

Fix: two new Linux-only helpers `SafeMkdirAll` and `SafeRemove` in `agent/exec/jail_linux.go`. Both walk path components via `openat2(RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS)`. A symlink anywhere in the chain surfaces as `ErrPathEscape` via the kernel's `EXDEV`/`ELOOP`. Non-Linux stubs return `ErrLandlockUnavailable` (unreachable in production because `configureJail` refuses-to-start without Landlock).

`configureJail`'s `WriteOpener` and `Remover` now route through these so the "no out-of-jail filesystem mutations" claim for in-process tools holds end-to-end.

### Refuse unsupported doublestar shapes (Copilot jail.go:101)

`ValidateWritablePaths` accepted any pattern that `path.Match` parses (after the `**` → `x` probe), but `matchOneGlob` only honors a documented set: bare `**`, `prefix/**`, `**/suffix`, `prefix/**/suffix`. Entries like `foo/**bar` or multiple `**` segments would silently never match at runtime — silent misapplication of a security boundary.

Fix: new `validateDoubleStarPlacement` helper splits the glob on `/` and requires every `**` to be its own segment, plus rejects more than one `**` segment per glob. Error message enumerates the supported shapes.

### Rebase

PR #273 (validation_overridden, Gap 5.2) merged to main during this review cycle. Two conflict resolutions:
- `cmd/tracker/main.go` — kept both imports (`execpkg` for `__jail-exec` dispatch + `pipeline` for `ErrValidationOverridden`).
- `CHANGELOG.md` — kept both `[Unreleased]` entries (writable_paths fs-jail block first, then validation_overridden). Both PRs remain unreleased.

Also updated the CHANGELOG entry's library-API delta to mention the new `RemoveFile` method on the `ExecutionEnvironment` interface and the `Remover` function-field on `LocalEnvironment` (added in `d19897a` as part of the codex P1 fix).

## Verification

- `go build ./...` clean on Linux + `GOOS=darwin GOARCH=amd64`
- `go vet ./...` clean
- `gofmt -l .` empty
- `go test ./... -short` — all 18 packages PASS
- `go test ./pipeline/handlers ./agent/exec -race` clean

CI should now be green again.

clintecker:
Round 4 pushed (`e127bd7`). Both new Copilot items addressed.

## Addressed

**Symlink-aware `working_dir` escape check** (Copilot jail.go:56)
Real working_dir relocation attack — string-only `Clean/Join/HasPrefix` math doesn't notice that `working_dir: link` (where `link → /etc`) resolves under cwd. After the string-cleaned containment passes, `validateWorkingDirEscape` now calls `filepath.EvalSymlinks` on both the resolved working_dir and the cleaned cwd, then re-checks containment against the real targets. `ErrNotExist` on the resolved path is the safe case — no symlink can have been placed at a path that doesn't exist yet.

**Refuse all brace patterns in `writable_paths` entries** (Copilot jail.go:95)
The matcher (`matchOneGlob` / `path.Match`) doesn't expand `{a,b}`. The old `balancedBraces` check accepted `workspace/*.{md,yaml}` as well-formed but then silently never matched anything at runtime — exactly the kind of surprising denial a security boundary shouldn't have. Replaced with `strings.ContainsAny(g, "{}")` rejection. Operators get an actionable hint: enumerate as separate comma-delimited entries (`workspace/*.md,workspace/*.yaml`). Removed the now-unused `balancedBraces` helper.

## Verification
- `go build ./...` clean
- `go vet ./...` clean
- `go test ./... -short` — all 18 packages PASS
- `TestValidateWritablePaths` (including the `malformed_brace_glob_is_rejected` row) still passes — the brace rejection is broader, not narrower.

clintecker:
Round 5 pushed (`28f8b83`). Addressed.

**`jailedWorkDir` fallback when `SessionConfig.WorkingDir` is empty** (Copilot backend_native.go:62)

Previously `filepath.Join(processCwd, "")` returned `processCwd` verbatim, silently relocating the jailed env to tracker's cwd whenever a caller set `AgentRunConfig.WorkingDir` (or relied on the backend env's WorkingDir) without explicitly populating `SessionConfig.WorkingDir`.

Resolution order is now:
1. `sessionCfg.WorkingDir` (per-node override)
2. `b.env.WorkingDir()` (the effective working dir the session would otherwise have used)
3. `processCwd` (last-resort fallback)

After resolving, `sessionCfg.WorkingDir` is set to the resolved path so `configureJail` validates against the same anchor the env is rooted at — without that sync step, an empty `SessionConfig.WorkingDir` would make configureJail anchor on `processCwd` while the env ran from the (potentially different) backend WorkingDir.

**Verification:** `go build ./...` clean, `gofmt -l .` empty, `go test ./pipeline/handlers ./agent/exec -short` green.

clintecker:
Round 6 pushed (`c3bb196`). Four addressed + two explicit pushbacks.

## Addressed

**SafeMkdirAll EEXIST race** (coderabbitai jail_linux.go:291) — concurrent creators between the ENOENT openat2 and the mkdirat now fall through to the re-open path instead of erroring. The re-open keeps RESOLVE_NO_SYMLINKS so a symlink planted by the racing creator is rejected, not followed.

**validateWorkingDirEscape walks deepest existing ancestor** (coderabbitai jail.go:68) — the round-4 fix returned nil on ErrNotExist for the whole path, which hid intermediate symlinks like `link/new` (where `link → /etc` exists but `new` doesn't). New `validateExistingAncestorEscape` walks up from the missing leaf to the deepest existing ancestor, EvalSymlinks-resolves it, and re-checks containment against the real processCwd.

**Glob normalization** (Copilot codergen_jail.go:71) — `ValidateWritablePaths` cleans entries before its checks, but the runtime matcher and Landlock dir computation consume the stored slice literally. `./workspace/**` validated yet then failed-closed because the matcher compared literal `./workspace` against `workspace/...`. Now `path.Clean`'d into the slice configureJail captures.

**Spec doc stale signature** (coderabbitai design.md:234) — `configureJail(cfg, attrs, env)` → `configureJail(cfg, env, processCwd)`. Matches the round-3 plan-doc fix.

## Explicit pushbacks

**coderabbitai jail_linux.go:148-170 (Landlock RWDirs "over-broad")**
The reviewer asks to either remove broad `RWDirs` and enforce per-file globs in userspace before exec, or restructure the allowlist so each writable item maps to a unique directory. **Both options reject what the spec explicitly settled in D2.** The two-tier semantic — in-process tools enforce *exact* globs via openat2; the Bash subprocess is bounded at the *directory ancestor* of each glob's static prefix via Landlock — is the documented behavior, called out in the CHANGELOG operator notes ("Bash subprocess is bounded at the **directory ancestors** of each glob's static prefix"). Landlock LSM is a directory-prefix kernel mechanism by construction; it does not and cannot do filename-level enforcement at the syscall layer. Userspace pre-validation of paths bash will touch is undecidable in general (bash can compute target paths at runtime from environment, variable expansion, command substitution). Operators who need file-scoped Bash bounding strip Bash from the agent's `tool_access`; that's documented in the CLAUDE.md gotcha section. Not changing.

**Copilot parent_death_linux.go (LockOSThread "dead weight")**
The `runtime.LockOSThread()` pairing is the deliberate fix for `PR_SET_PDEATHSIG`'s thread-scoped delivery semantics, called out by the **earlier** coderabbitai round-2 review (parent_death_linux.go:43) and shipped in `d19897a`. Without it, the Go runtime can migrate the calling goroutine off its original OS thread and the thread is later retired during GC stop-the-world or thread-pool reduction — at which point the kernel sends a premature SIGKILL to the child even though the parent process is still alive. The lock holds the thread alive for the lifetime of `cmd.Run()`, which is exactly the window where Pdeathsig must remain armed against the spawning thread. Removing it would re-open the premature-kill hole. Not changing.

**macOS/BSD enforcement**
The PR ships writable_paths as Linux-only with a refuse-to-start gate on non-Linux (G3 in configureJail; `jail_other.go` is a deliberate stub). Making it actually enforce on macOS/BSD is per-OS feature work (FreeBSD Capsicum, macOS Sandbox) that is its own multi-week project. Not in scope for #272.

## Verification

`go build ./...` clean on Linux + `GOOS=darwin GOARCH=amd64`. `go vet ./...` clean. `gofmt -l .` empty. `go test ./... -short` green.

clintecker:
Follow-up issues filed for the two pushbacks that warrant tracked future work:

- **#280** — *writable_paths: investigate file-scoped Bash enforcement.* Covers the spec D2 two-tier semantic, the Landlock directory-prefix kernel constraint, and four options (refuse-to-start on file-scoped + bash combo, doctor warning, per-file workspace mapping, louder docs). Recommends options 1 + 4 first; defers 2 and 3 until adopter feedback.
- **#281** — *writable_paths: per-OS enforcement on macOS (Sandbox) / FreeBSD (Capsicum).* Each is a 1-2 week project per platform with its own openat2-equivalent in-process tier. Documents what the work looks like; priority-bumped only when a real adopter asks.

The Pdeathsig `LockOSThread` pushback (Copilot `parent_death_linux.go`) does not get a follow-up issue — the lock is the **correct shipping fix** for the thread-scoped SIGKILL hole flagged by coderabbitai round 2 (`d19897a`). Filing an issue against working code would be noise.

clintecker:
Round 7 pushed (`6a1c616`). Both Copilot findings addressed.

## SECURITY: writable_paths gate bypassed for claude-code/acp (codergen.go:647)

Real correctness gap. `configureJail`'s G2 backend gate only fires inside `NativeBackend.Run`. When the selected backend is `claude-code` or `acp`, `buildRunConfig` switches `Extra` away from `*agent.SessionConfig` before `backend.Run` is called — the `WritablePathsSet` signal never reaches the gate. A node that declared `writable_paths` but selected (or inherited via `--backend`) claude-code/acp would silently start **unjailed** instead of the spec-promised refuse-to-start.

Fix: new `refuseWritablePathsOnUnsupportedBackend` gate runs in `CodergenHandler.Execute` right after `selectBackend`. The check is by type assertion against `*NativeBackend` — anything else cannot run inside the jail, regardless of how it got selected (node attr, `--backend` default, etc.). Tested with a fake backend stand-in across nil-node / no-attr / writable+native / writable+non-native rows.

## CommandWrapper in-place mutation contract (local.go:178, :344)

CommandWrapper was previously allowed to return a different `*exec.Cmd`. A wrapper that did so would silently drop `Cancel`, `WaitDelay`, and `SysProcAttr` (including `Pdeathsig` + `Setpgid`) — re-opening the very orphan-leak window `LockOSThread` + Pdeathsig are meant to close.

Fix: enforce pointer identity at both `ExecCommand` and `ExecCommandWithLimit`. Return a clear error if the wrapper hands back a different pointer. The jail's `WrapBashCmd` already mutates in place so no consumer changes. Re-applying the standard fields on the returned cmd was considered and rejected — that would re-introduce the exact orphan-leak window the contract is meant to preserve.

## Verification

`go test ./... -short` green (all 18 packages); `gofmt -l .` empty.

clintecker:
Round 8 pushed (`f4a2a4e`). All five Copilot items addressed.

## Code fixes

**`RESOLVE_NO_MAGICLINKS` on `SafeMkdirAll` + `SafeRemove`** (Copilot jail_linux.go:282, :346)
`OpenForWrite` has used `RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS` since round 3. The follow-up symlink-safe helpers added BENEATH + NO_SYMLINKS but omitted NO_MAGICLINKS — leaving a procfs magic-link edge for the dir-creation and delete paths that the in-process write seam already closes. Same Resolve flags now on both helpers, with doc comments updated in lockstep.

## Documentation corrections

**Accurate Pdeathsig semantics** (Copilot parent_death_linux.go:30, local.go:198)
You're right — the earlier coderabbitai round-2 review cited a web-query result claiming PDEATHSIG is per-thread, which matched older kernel behavior but is contradicted by current Linux: `prctl(2)` delivers on parent PROCESS exit. Both comments are reworded:
- `parent_death_linux.go` now explicitly states the modern semantics, acknowledges the older-kernel scenario as defensive depth, and lists three concrete reasons `LockOSThread` stays (defense in depth on unknown kernel versions, stable thread identity for future per-thread state, negligible cost given serial per-agent invocation).
- `local.go` points at `parent_death_linux.go` for the full rationale.

Keeping `LockOSThread` per your own note in the review ("can still be kept as a conservative measure if desired"). Removing defensive code that's caught no bug would be premature.

**CLAUDE.md brace wording** (Copilot CLAUDE.md:376)
The refuse-to-start list said "unbalanced braces" but the validator rejects **any** brace usage (round 4 fix). Updated to "any brace usage" and expanded the list to cover the other refuse-to-start classes that landed in later rounds: symlink-aware `working_dir`, unsupported doublestar shapes, malformed character classes, and the dispatcher-layer non-native-backend gate from round 7.

## Verification

`go test ./... -short` green; `gofmt -l .` empty; `go vet ./...` clean.

clintecker:
Round 9 pushed (`522cc39`). Both test-reliability items addressed.

## Forge-side vacuity guard on TestParallelBranchSymlinkRace (Copilot e2e:325)

Round 3 added a guard on branch B's WriteFile success count — catches "every in-jail write was rejected" vacuity. But branch A's forge loop discards ExecCommand's error, so a broken forge path (missing `sh`/`ln`, `__jail-exec` dispatch broken, Bash jail too tight) leaves branch B writing to the static pre-created directory and the race is never actually raced. The "outsideDir is empty" assertion would then pass without exercising RESOLVE_NO_SYMLINKS at all.

Fix: `sync/atomic` counter on the forge goroutine — when `ExecCommand` returns nil error AND `ExitCode == 0`, increment. After the loop joins, require ≥1 successful forge. The kernel-rejection assertion on `outsideDir` then provably reflects the race being exercised, not just branch A silently failing to ever create the symlink.

## Broaden Landlock skip in TestJailExecDispatch (Copilot jail_exec_test.go:18)

The previous guard only skipped on `errors.Is(probeErr, ErrLandlockUnavailable)`. Restricted CI environments can fail `ProbeLandlock` with EPERM, seccomp errors, or other shapes that don't wrap the sentinel — those would propagate and force a test failure even though Landlock enforcement couldn't be exercised.

Fix: skip on any non-nil `ProbeLandlock` error. Matches the broader-skip pattern used by every other Landlock-gated test in this PR.

## Verification

`go test ./... -short` green; `gofmt -l .` empty; race-detector clean on the rewritten test.

clintecker:
Round 10 pushed (`6c9299b`). All three Copilot items addressed.

## SECURITY/correctness: validation root is the session, not the process cwd (backend_native.go:78)

Real issue. `ValidateWritablePaths`'s containment check was rooted at `os.Getwd()` — whichever directory the user happened to be in when they ran `tracker`. If invoked as `cd /tmp && tracker --workdir /projects/foo`:

- A node-level absolute `working_dir` outside `/tmp` could pass validation when it should have been checked against `/projects/foo` (escape direction).
- A perfectly valid relative `working_dir` resolved against `/tmp` instead of `/projects/foo` (silent relocation away from the session).

Fix: use the backend `LocalEnvironment.WorkingDir()` as the session root for both relative-resolution AND `configureJail`'s containment base. Process cwd is now a fallback only when the env has no WorkingDir set. The "session root" is the resolved `--workdir` flag (or `AgentRunConfig.WorkingDir`), which is the actual session boundary.

## Pre-create Landlock RW roots through the anchor (jail_linux.go:151)

Real ergonomic gap. `landlock.V3.RestrictPaths` refuses to add a rule for a non-existent directory, and the jailed shell can't create the root itself because its parent (the anchor) is read-only under `RODirs("/")`. So flows like `mkdir -p workspace && go build` failed under `writable_paths` whenever the workspace dir wasn't pre-staged before tracker started.

Fix: `RunJailExec` now walks each rwDir relative to the anchor and calls `SafeMkdirAll` before `RestrictPaths`. `SafeMkdirAll` uses `openat2(RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS)` on every component so a pre-existing symlink at the rwDir path can't redirect the pre-creation outside the anchor. The bare-`**` case (rwDir == anchor) skips since the anchor already exists by construction.

## Docstring fix (codergen_jail.go:62)

Updated `configureJail`'s docstring: called from `NativeBackend.Run` (not `buildConfig`) with a fresh `*LocalEnvironment`. Refusals surface as the `SessionResult` error which `Execute` turns into `EventNodeFailed` pre-LLM-token. Added a note that claude-code/acp never reach this function — they're refused earlier at `refuseWritablePathsOnUnsupportedBackend` (round 7) because `buildRunConfig` drops the SessionConfig signal for those backends.

## Verification

`go test ./... -short` green; `gofmt -l .` empty.

clintecker:
Round 11 pushed (`81b77d2`). Both Copilot items addressed — these are real glob-validation gaps with security/correctness implications.

## Reject any `..` segment fail-closed (jail.go:162)

The existing parent-escape check examined `filepath.Clean(g)` for `..` or `../`-prefixed forms — fine for outward escapes but blind to **inward collapses** like `workspace/../**` which cleans to `**`. `configureJail`'s round-6 `path.Clean` normalization (added to fix the `./workspace/**` literal-prefix mismatch) then stores `**`, silently granting writes across the entire anchor instead of the workspace subtree the author probably tried to constrain themselves to. Real permission escalation via cleverly constructed input.

Fix: reject any glob entry containing a `..` segment, before any `Clean` is applied. Error message is explicit about why: "rejected fail-closed because path.Clean would collapse it to a broader glob than authored."

## Reject glob metachars in any segment before `**` (jail.go:225)

`validateDoubleStarPlacement` previously only checked that `**` was its own segment and that there was at most one of them. But the prefix part (segments before `/**`) is taken LITERALLY by `matchOneGlob` (`strings.HasPrefix`), while `landlockDirForGlob` derives the RWDir from the static prefix BEFORE the first metachar.

For `work*/**`:
- `matchOneGlob`.prefix = `"work*"`; `relPath` like `"workspace/file"` never starts with literal `"work*/"` → matchOneGlob always denies.
- `landlockDirForGlob`.static-prefix = `"work"` → `filepath.Dir("work")` = `.` → collapses to the anchor itself → Landlock grants write on the **ENTIRE anchor**.

That's broader than authored permissions for Bash AND silently empty for in-process tools — both tiers disagree with each other and with what the operator wrote.

Fix: scan the segments before the `**` index and refuse any that contain `*`, `?`, or `[`. Error message names the offending segment and explains both halves of the mismatch.

## Verification

`go test ./... -short` green; `gofmt -l .` empty.

clintecker:
Round 12 pushed (`4c241c6`). All five Copilot items addressed.

## `validateWorkingDirEscape` resolves processCwd symlinks first (jail.go:67)

On distros where `/home` is a symlink to `/var/home` (or similar setups where tracker is invoked from a symlinked path), the string-cleaned containment check rejected perfectly valid `working_dir` values expressed as the real path. `EvalSymlinks` ran AFTER the string check, so the reject already happened.

Fix: resolve processCwd's own symlinks up front. The containment check now passes if working_dir is contained under EITHER the cleaned cwd OR the symlink-resolved cwd. EvalSymlinks falls back to the cleaned form on error so an unrelated FS issue doesn't fail-close on the operator.

## EACCES no longer misclassified as `ErrPathEscape` (jail_linux.go:113, 322, 383)

You're right — `openat2` returns EACCES for ordinary permission issues (directory mode/ownership) inside the anchor, not just resolve-escape signals. Wrapping it as `ErrPathEscape` made callers think the agent was trying to escape when the real failure was a vanilla permission problem on a parent dir. Affected `OpenForWrite`, `SafeMkdirAll`, and `SafeRemove`.

Fix: dropped EACCES from the EXDEV/ELOOP case branch in all three. `ErrPathEscape` now fires only on the kernel's actual resolve-escape errnos: EXDEV from `RESOLVE_BENEATH`, ELOOP from `RESOLVE_NO_SYMLINKS` / `NO_MAGICLINKS`. EACCES propagates as a regular wrapped error so `errors.Is` classification upstream stays correct.

## Short writes + Close error propagation in `WriteOpener` path (local.go:125)

The jailed write path did `defer f.Close()` + `_, err = f.Write(...)`. That ignored Write's `n` return (silent truncation on partial writes) and swallowed Close's error (NFS commit, fsync failures, delayed I/O — all lost). `os.WriteFile` in the unjailed path handles both internally; the jailed path silently regressed.

Fix: new `writeViaOpener` helper does the canonical Go pattern:
- Check `n < len(data)` → `io.ErrShortWrite`
- `defer` Close with `if err == nil { err = cerr }` so Close errors surface when the write itself succeeded
- Write errors win over Close errors when both fire

## Verification

`go test ./... -short` green; `gofmt -l .` empty.

clintecker:
**Sweep pass** (`8bd283a` + `88463c6`) — stepping back to generalize the patterns we kept finding round-by-round, then auditing the tree for unfound instances.

## Pattern taxonomy (rounds 1-12)

1. String-based path checks blind to symlinks → rounds 3, 4, 10, 12
2. Validators accepting shapes the matcher won't honor → rounds 4, 5, 11
3. EACCES misclassified as ErrPathEscape → round 12 (3 sites)
4. Refuse-to-start gates firing too deep → round 7 dispatcher gate
5. Vacuous tests when subprocess silently fails → rounds 3, 9
6. Short writes / Close / ExecCommand errors dropped → round 12
7. Optional-hook contracts not runtime-enforced → round 7 CommandWrapper
8. Documentation lag on signatures and semantics → rounds 6, 8, 10
9. Setup assumes state that doesn't exist → round 10 Landlock pre-create

## New finding — SECURITY: GenerateCodeTool + WriteEnrichedSprintTool bypass jail

Audit hit. **Same shape as the round-1 apply_patch fix, just in different tools.** Both LLM-invokable tools called `os.MkdirAll` + `os.WriteFile` directly. A jailed session with `TRACKER_CODEGEN_MODEL` or `TRACKER_SPRINT_WRITER_MODEL` configured could let the cheap-model output land anywhere on disk.

Fix in `8bd283a`:
- Both tools now take optional `WithGenerateEnv` / `WithSprintWriterEnv` options.
- Internal write helpers (now methods on the tool) prefer the env when set: `filepath.Rel` against `env.WorkingDir()`, then `env.WriteFile(ctx, rel, content)` — routed through the jail's `WriteOpener` → `openat2` + glob check.
- `backend_native.go` registers both tools with the resolved `env` variable, which is the **jailed** env when `sessionCfg.WritablePathsSet`.

Same writable_paths enforcement now applies whether the agent uses `apply_patch`/`Write`/`Edit` OR `generate_code`/`write_enriched_sprint`.

(`DispatchSprintsTool` only does `os.Open` for reading + stderr writes; reads are allowed everywhere by `RODirs("/")` and stderr is not jail-relevant — no changes there.)

## New finding — Test coverage gaps for validator refusals

Eleven new rows on `TestValidateWritablePaths` in `88463c6` exercising every refusal class introduced across rounds 4-11 (balanced braces, `..` segments, metachars before `**`, multiple `**`, malformed char classes, Windows abs) plus happy-path coverage for every supported doublestar shape. Each asserts the actual error substring so future behavior regressions fail loudly.

## Audited and OK

- All other `pipeline.AgentBackend` implementations: `ACPBackend.Run`, `ClaudeCodeBackend.Run` already gated by the round-7 dispatcher `refuseWritablePathsOnUnsupportedBackend`. They never reach `configureJail`.
- Other `os.Remove`/`os.MkdirAll`/`os.WriteFile` in `agent/tools/`: only the two found above were write-side. `dispatch_sprints.go` is read-only.
- `agent/exec/local.go` `safePath` is the string-based pre-check; downstream `WriteOpener`/`openat2` catches what string ops miss, and tests / direct ReadFile callers use it as documented.
- Remaining `ErrPathEscape` sites: all EXDEV/ELOOP only.
- Remaining `ExecCommand` error sites: all propagate; no production-code `_, _ = env.ExecCommand` discard patterns.

`go test ./... -short` green; `gofmt -l .` empty; `go vet ./...` clean.

clintecker:
Round 13 pushed (`f125c90`). All three Copilot items addressed.

## `WithGenerateEnv` / `WithSprintWriterEnv` now default workDir from env (Copilot generate_code.go:51 + write_enriched_sprint.go:57)

Real DX gap from the round-`8bd283a` audit fix. A caller that supplied env but not `WithGenerateWorkDir` got the tool with empty `t.workDir` but non-nil `t.env`. `Execute` then resolved `output_dir` relative to "." (the empty workDir), producing absolute paths under processCwd — which may not be `env.WorkingDir()` at all. `writeFile` then computed `filepath.Rel(env.WorkingDir(), absPath)` yielding leading `..` segments, and `env.WriteFile` rejected fail-closed. Net result: the tool silently stops writing files.

Fix: both `WithGenerateEnv` and `WithSprintWriterEnv` now default `t.workDir` to `env.WorkingDir()` when `t.workDir` is still empty. `backend_native.go`'s registration sets both explicitly so its behavior is unchanged; the default only fires for callers that construct the tool with just env (tests, embedded integrations).

## Accurate `ValidateWritablePaths` error-sentinel docstring (Copilot jail.go:26)

The docstring claimed "class-1 and class-3 wrap `ErrPathEscape`; class-2 is plain." But `validateGlobEntry`'s malformed-pattern paths (brace usage, multiple `**`, metachars before `**`, glued `foo/**bar`, unbalanced character classes) return plain errors — those are "bad glob shape" not "escape attempt."

Reworked to spell out exactly which class-3 subclasses get `ErrPathEscape`-wrapped (the escape-flavored ones — absolute, `~`, parent traversal, Windows-absolute, inward `..` segments, non-existent intermediate symlinks) versus which are plain (the malformed-shape ones). The codergen handler refuses to start on either; the sentinel distinction is for upstream consumers that want to differentiate "operator wrote a bad pattern" from "attacker tried to point the jail outside the session root."

## Verification

`go test ./... -short` green; `gofmt -l .` empty.

clintecker:
Filed five follow-up issues capturing what I should have been doing across this PR's lifetime. These are mine to own; flagging them so the post-merge work is tracked, not lost.

| # | Issue | Why |
|---|---|---|
| [#282](https://github.com/2389-research/tracker/issues/282) | Property/invariant tests for the jail's public surface | The contracts I kept getting wrong on (path containment, glob shapes, sentinel wrapping, RESOLVE flags) should be pinned by `testing/quick` properties, not example tests. |
| [#283](https://github.com/2389-research/tracker/issues/283) | Threat-model doc + checklist for the LLM tool write surface | The `generate_code` / `sprint_writer` bypass had been latent for months — same shape as round-1's `apply_patch` fix. A checklist + CI lint (no `os.WriteFile` in `agent/tools/`) prevents the next instance. |
| [#284](https://github.com/2389-research/tracker/issues/284) | Linux security primitives reference doc | `PR_SET_PDEATHSIG` modern vs legacy, `EACCES` vs `EXDEV`/`ELOOP`, `RESOLVE_*` flag interactions, Landlock directory-prefix vs glob — I re-derived (and once got wrong) across rounds. Should be documented once for the next contributor. |
| [#285](https://github.com/2389-research/tracker/issues/285) | 9-class audit checklist for writable_paths changes | The taxonomy from `8bd283a`'s sweep, formalized as a PR-template item so future jail-touching changes get checked against each class before merge. |
| [#286](https://github.com/2389-research/tracker/issues/286) | "Freeze and prove" process pattern for security PRs | The meta-process fix: spec → API freeze → property tests → small implementation → checklist sweep. CLAUDE.md addition so it's institutional. |

Ready for merge whenever you are.

PR #317, opened 2026-06-05
feat(engine): turn-limit breach = guard not guillotine — verify → classify (#303 PR1)

clintecker:
## Summary

**PR1 of 2** for #303 (Phase-1 turn-limit track; builds on merged #302/#295/#297). Refs epic #308.

Turn-limit exhaustion no longer maps unconditionally to `OutcomeFail`. On a **native-backend** breach the engine now runs a verify pass and *classifies* the breach instead of guillotining it:

- **verified-green** → `OutcomeSuccess`, advancing through the pipeline's own commit-on-success node (build_product's `CommitIfDirty`, #297). This alone would have saved `code-goblin` run `7b6e08c9e2b2` (green at turn 48 but discarded).
- **LoopDetected** → pathological, `OutcomeFail` (stops, as today).
- **anything else** (verify red / not-run) → `OutcomeFail` + `ctx.turn_breach_class = operator_decision` for routing.
- `turn_breach_policy: fail` (declare under a `params:` block) opts back into today's guillotine, byte-for-byte.

PR2 will add the operator-decision `wait.human` node + warm `continue +N` and will close #303.

## The design was panel-reviewed first

A 6-expert panel reviewed the design before any code. It corrected the original premise and caught five real bugs the first draft would have shipped — all absorbed here:

1. **commit-if-green is classification-only.** The engine's git machinery is **off by default** (`WithGitArtifacts` is never wired in the CLI) and, even when on, commits the *artifact* dir — not the agent's working tree. So persistence is the pipeline's job (build_product's `CommitIfDirty`); PR1 adds **zero** engine commit code. Verified at the routing level (`TestBuildProductIssue303GreenBreachRescuePath`), not via the artifact repo.
2. **`auto_status` can no longer manufacture a breach success.** `parseAutoStatus` defaults to success on a missing `STATUS:` line (the normal outcome of running out of turns); applying it on a breach defeated the "never silent-advance" invariant. It's now honored only on normal completion. (Updated `TestCodergenHandlerMaxTurnsWithAutoStatus_BreachIgnoresAutoStatus` to pin the new safe contract.)
3. **The verify-on-breach *decision* lives in codergen, not the agent.** codergen sets `SessionConfig.VerifyOnBreach = (policy != "fail")`; the agent is mechanism-only, so the opt-out path pays no verify cost.
4. **Native-only is an explicit guard**, not an accident of which backend populates `MaxTurnsUsed`.
[...]

Issue #313, opened 2026-06-05
engine: parallel branches bypass strict-failure routing + aggregateStatus masks single-branch failure


clintecker:
## Problem

Discovered during review of #296 / PR #312 (Codex P1).

Two related engine-level gaps in the parallel handler (`pipeline/handlers/parallel.go`) mean a `.dip` workflow cannot fully route the failure of an agent node that runs as a **parallel branch**:

1. **Per-branch `fallback_target` is runtime-inert.** `ParallelHandler.runBranch` (~L270) calls `h.registry.Execute(execCtx, tn, branchCtx)` directly and records a `ParallelResult`. It never enters the engine run loop, so `Engine.checkStrictFailure` / `findFallbackTarget` (`engine.go:525`, `engine_checkpoint.go:217`) never apply to branch nodes. A `fallback_target` (or any failure edge) declared on a branch node does nothing at runtime.

2. **`aggregateStatus` is success-if-any, masking single-branch failure.** `aggregateStatus` (~L324) returns `OutcomeSuccess` if **at least one** branch succeeded. In `build_product.dip`, if the adversarial `ReviewGemini` exhausts `max_turns` and fails but `ReviewClaude`/`ReviewCodex` succeed, `ReviewParallel` aggregates to success and proceeds to `ReviewJoin` → `SynthesizeReviews` with that review **silently missing**. Nothing downstream detects the gap: `SynthesizeReviews` has no missing-file guard, and `FinalSpecCheck` explicitly does **not** backstop the reviewer-rubric smell classes (W4/W5/W13). This defeats the cross-review safety net #233 built.

## What #296 / PR #312 did (and didn't)

PR #312 handled the **main-loop** nodes (`Implement`, `ApplyReviewFixes`, `FinalCommit`) with `fallback_target`, and routed the **all-three-reviewers-fail** case via a conditional fail edge on the aggregating `ReviewParallel` node (which IS a main-loop node and previously dead-stopped on all-fail). It removed the inert per-branch `fallback_target` attrs rather than ship a lie. The **single-branch-masked** case cannot be fixed in `.dip` and is deferred here.

## Proposed fix (engine, Go)

- Thread per-branch `fallback_target` through `runBranch`, OR make branch failure observable to the engine's strict-failure path.
- Make the fan-in aggregation policy configurable instead of blanket success-if-any — e.g. a `quorum` / `all`-success / required-branch policy so an adversarial reviewer's failure is not maskable.

## Context

[...]

Issue #310, opened 2026-06-04
adapter+engine: consume dippin-lang budget/limit attrs (stall_timeout, graph budget ceilings, max_turns exhaustion) — upstream of dippin-lang#94 (PR #96), unblocks #67


clintecker:
## Problem
dippin-lang#94 (shipped in dippin **PR #96**) adds declarable budget/limit attrs to `.dip`. Closed #67 was blocked precisely on these fields not existing in `ir.WorkflowDefaults`. Now that they exist, tracker must consume them so the budget-by-cost + no-progress redesign (#304) and the turn-limit guard (#303) can be configured from the workflow file instead of only CLI flags.

> **Note (2026-06-05):** this issue was originally drafted from dippin's pre-implementation sketch. The body below has been reconciled to the **actual** surface dippin shipped in PR #96 — the field set, naming, and `max_turns` semantics changed during implementation. See the reconciliation comment.

## What dippin actually shipped (PR #96)
**Graph-level `defaults` only (all four; no per-node budgets):**
- `max_total_tokens` — int, hard ceiling. `0`/unset = no limit.
- `max_cost_cents` — int, **US cents** (`1000` = $10.00). `0`/unset = no limit.
- `max_wall_time` — duration, wall-clock ceiling. `0`/unset = no limit.
- `stall_timeout` — duration; abort/route when no forward progress for that wall-clock span. `0`/unset = disabled. **(This is the field formerly sketched as `no_progress_window`; it is a wall-clock `time.Duration`, NOT an iteration count. New key — not in #67.)**

**Per-node:** only the pre-existing `max_turns` (int). dippin **deliberately deferred** all per-node budget ceilings (`max_cost_usd`, per-node stall, etc.) as YAGNI — they do not exist. Don't build against them.

**`max_turns` exhaustion semantics (the new contract):** reaching `max_turns` ends the node with outcome **`fail`** — there is **no declarable action enum** (`fail`/`fallback`/`truncate` was declined as YAGNI). The failure routes through the **same graph-level `on_failure` cascade** as any other agent failure (fail edge → bounded retry → `fallback_target` → graph `on_failure` → halt). `stall_timeout` and budget-ceiling breaches route through the **same** cascade. This is the #309 / engine-#295 machinery — this issue's engine work is "treat budget/stall/turn breach as a routable failure," not "honor a per-node action field."

[...]

Issue #308, opened 2026-06-04
Epic: build_product hardening — structural & process gaps beyond #233 (case study: code-goblin run 7b6e08c9e2b2)


clintecker:
# build_product hardening — the next generation of gaps

Issue #233 hardened `build_product` against **content** deviations (spec-literal grep, interface reachability, test-verifies-contract, scope-creep, evidence-weighted synthesis, adversarial review). That machinery is good and stays. This epic tracks the **structural / process** gaps it didn't cover — the ones that made a real run fail and ship bugs.

## The case study
A `build_product` run (`code-goblin`, run `7b6e08c9e2b2`) failed. Proximate cause: the `Implement` agent node (`max_turns: 50`) exhausted its turn budget on the final milestone. The work was GREEN (`go test ./...` + `go vet` passed at turn 48) but **uncommitted** — the agent spent its last turns re-verifying instead of committing. Turn-exhaustion → `OutcomeFail`; `Implement` has no fail edge and no `fallback_target`, so the engine's strict-failure rule **halted the entire pipeline**. The cross-review phase (incl. adversarial ReviewGemini) and FinalSpecCheck — the safety net designed to catch exactly the bugs that shipped (a `rand.Int63n(0)` panic on `Retry-After: 0`, an off-by-one retry count) — **never ran**.

A full audit of what it built surfaced several classes of miss that are universal (spec-agnostic, language-agnostic):
- unhandled agent failure silently halts the whole pipeline;
- green-but-uncommitted work is lost on node death;
- ~20% of the turn budget is burned re-reading context every node (per-node amnesia);
- quality gates only fire when a Makefile exists (no lint/complexity for plain Go/JS/Rust/Python repos);
- build/test detection is first-match (polyglot repos under-tested);
- spec defects (dangling refs, contradictory constants, prose guarantees, seams that can't satisfy their contracts, mandated tests owned by no milestone) flow straight into the build.

### Phase 0 — Stop the bleeding (P0)
The silent-halt bug, fixed at both layers. Ship first; it's what lets every other safety net actually run.
- [x] #295 — engine: graph-level fallback catch-all for unhandled agent failure
- [x] #296 — build_product: failure/exhaustion routing on every agent node

### Phase 1 — Never lose work, never overrun blindly (P1)
- [x] #302 — engine: commit-WIP to a recoverable ref before routing a failed node
- [x] #297 — build_product: commit-first / stop-when-green + commit-if-dirty node
[...]

PR #330, opened 2026-06-09
feat(turn-guard): operator-decision node + warm continue+N (#318)

### LINKED ISSUE #318: engine: operator-decision node for steady-progress turn-breaches + warm continue+N (#303 PR2)
## Context

Follow-up to **#303** (closed; PR1 #317 merged). #303 PR1 delivered the engine/agent **classification core** of the turn-limit guard: a native-backend breach now runs a verify pass and classifies — verified-green → `OutcomeSuccess` (advances through the pipeline's commit-on-success node), `LoopDetected` → pathological `OutcomeFail`, anything else → `OutcomeFail` + `ctx.turn_breach_class = operator_decision`. A `turn_breach_policy: fail` opt-out reproduces the old guillotine.

PR1 deliberately split off the **operator-decision node + warm `continue +N`** as a second PR. #303 was auto-closed by the PR1 merge before that second half landed, so this issue tracks the remaining, already-designed work. Refs epic #308 (Phase 1).

**Design already written:** `docs/superpowers/specs/2026-06-05-issue-303-turn-limit-guard-design.md` (see the "PR2 — operator-decision node + warm continue" section). The hazards below were surfaced by a 6-expert panel review during #303 brainstorming.

## Scope

When a breach classifies to `operator_decision` (steady progress, non-green), give the operator a real choice instead of a bare fail/fallback, and make unattended runs take a safe default deterministically.

1. **Operator-decision node = reuse `wait.human`** (freeform mode, no new handler), reached via an author-wired conditional edge `<agent> -> <Operator> when ctx.turn_breach_class = operator_decision`. Labeled edges: `continue` / `commit_advance` / `stop` / `abandon`.
2. **Warm `continue +N`** re-enters the agent node with a bumped `MaxTurns` and carried `PriorEpisodeSummaries` (warm, not cold), bounded by a cap.
3. **Autonomous default** for unattended runs (`--auto-approve` / `--autopilot` / `--webhook-url`): default = **stop + commit-WIP + route-to-fallback**, never a silent advance past unverified work.
4. Wire the operator node into `examples/build_product.dip` and re-pass `dippin doctor` / `simulate -all-paths`.

## Acceptance criteria (the #303 ACs PR1 did not cover)

- The operator-decision node exposes the four options + a **configurable autonomous default**; unattended runs (`--auto-approve` / `--autopilot` / `--webhook-url`) take the default **deterministically**.
- `continue +N` resumes with **non-empty `PriorEpisodeSummaries`**, bounded by the cap.
- A green `commit_advance` and a `stop` both persist work and route safely; `abandon` does not advance.

## Design hazards to handle (from the panel review)

1. **Auto-approve default-ordering footgun (CRITICAL).** Freeform mode reads the `default` attr (NOT `default_choice`). If unset and `continue` is edge #1, `--auto-approve` *continues* — the exact unsafe outcome the issue forbids. Pin `default: stop`, order `stop`/`abandon` **before** `continue`, and add a guard test that the resolved unattended default is never `continue`. Note `--autopilot lax`/`mentor` bias to "forward progress" and are not deterministically safe — frame the prompt so "continue" is the risky option (or don't route this gate through the LLM judge).
2. **Warm-resume `MaxTurns` bump needs NEW plumbing.** `MaxTurns` is read statically in `buildConfig` (`node_config.go` → `codergen.buildConfig`); nothing reads context. Add a context/disk-driven `MaxTurns` override. Episodes already carry across (warm) for free via `ContextKeyEpisodeSummaries`.
3. **Cap circuit-breaker** belongs in a tool node gating the operator's `continue` edge (mirror build_product's `fix_attempts` disk counter), NOT the global engine `RestartCount` (shared across all loops, reset semantics differ). `BudgetGuard` covers only a global cost backstop, not a per-loop ceiling.
4. **`commitWIPBeforeRouting` bypass.** It runs only inside `checkStrictFailure`, which early-returns when a node has conditional edges — so adding the operator conditional edge means a fail→operator route skips WIP preservation. (Moot while `WithGitArtifacts` is off by default, but real if enabled.) Cover it.
5. **dippin:** any new node attr rides in a `params:` block (a bare attr is a fatal parse error in v0.35.0). The new cycle (`Operator -> <agent>` continue) + 4-way gate expands `simulate -all-paths`; every operator edge must be labeled and reach a terminal (DIP005).
6. **UX:** rely on the standard `last_response` append (`\n\n---\n`) so the gate routes through `ReviewHybridContent` (fullscreen glamour + radio labels); don't inline the long breach message into the node `label`. Ensure Cancellable/Ctrl-C.
7. **Windows:** any breach-verify entry point touched stays inside the `//go:build !windows` constraint (agent/ is already !windows-only in practice).

## Out of scope

#304 (cost + no-progress detector), #298/#299/#300/#301/#305/#306/#313.

clintecker:
Closes #318 — the deferred PR2 of #303 (epic #308, Phase 1 turn-limit track). PR1 (#317) merged the breach-classification core; this delivers the operator-decision node + warm `continue +N` it split off.

## What

When a **native-backend** turn-limit breach classifies to `operator_decision` (PR1: steady progress, verify ran but **not** green, no loop), the operator now gets a real choice instead of a bare escalation — and unattended runs take a deterministically **safe** default.

In `build_product.dip`, `Implement` routes `when ctx.turn_breach_class = operator_decision` to a new **`OperatorDecision`** gate. It **reuses `wait.human`** (freeform) — **no new handler** — with four labeled edges:

| label | route | meaning |
|---|---|---|
| `stop` *(default)* | `EscalateMilestone` | treat as stuck, preserve work, escalate |
| `abandon` | `Done` | end the run, do not advance |
| `commit_advance` | `CommitIfDirty` | persist the tree via the commit-on-success node, then advance |
| `continue` | `ContinueWithMoreTurns` | warm retry with a bumped turn budget (capped) |

## Hazards handled (from the #303 panel review)

1. **Safe unattended default (CRITICAL).** Pin `default: stop` and order `stop`/`abandon` **before** `continue`. `--auto-approve` / `--webhook-url` resolve to the freeform `labels[0]` fallback (= `stop`); the timeout path resolves `default_choice` (= `stop`). Either way an unattended run **never silently advances unverified work**. `--autopilot lax`/`mentor` are not deterministically safe, so the gate prompt frames `continue` as the risky option. A **pathological** breach still narrowly falls through to the `EscalateMilestone` catch-all.
2. **Warm `continue +N` plumbing.** New node-scoped, disk-driven `MaxTurns` override (`.tracker/turn_overrides/<nodeID>`) consulted in `codergen.buildConfig`. `ContinueWithMoreTurns` bumps `Implement`'s budget (50 → 90/130/170) and re-enters it **warm** — `PriorEpisodeSummaries` carry across for free. `resolveMaxTurns` keeps `buildConfig`'s branch count flat (no complexity regression).
3. **Cap circuit-breaker.** A per-loop disk counter (NOT the global engine `RestartCount`); 3 continues max, then escalate. Counters reset at the milestone boundary (`MarkMilestoneDone`) and in `Cleanup`.
4. **`commitWIPBeforeRouting` bypass.** Documented + tested: `Implement` already carried conditional edges, so the operator route introduces **no new** bypass (and it is a no-op while git-artifacts is off by default). Green work is persisted the way all product persistence works — through the pipeline's own commit node.
5. **dippin.** New cycle + 4-way gate; `dippin validate` clean, `doctor` **A/90** (baseline held), `simulate -all-paths` enumerates the operator cycle and **all paths terminate**.
6. **UX.** Relies on the standard `last_response` append (no inlined breach message in the label) so the gate renders via `ReviewHybridContent`.

## TDD

Failing tests written first, then implementation:
- `codergen_warmturns_test.go` — `buildConfig` honors the disk override (+ node-scoping, no-override defaults); warm continue carries bumped `MaxTurns` **and** non-empty `PriorEpisodeSummaries`.
- `build_product_operator_decision_test.go` — node exists / reached by the `operator_decision` condition; safe default + edge order; four labeled edges each terminate; capped warm-retry loop; pathological still falls through; persists via the commit node.
- `operator_decision_default_test.go` — behavioral guard: `AutoApprove` through `HumanHandler.Execute` resolves to `stop`; dropped-default fallback is never `continue`.

## Verification

- `go build ./...` + `GOOS=darwin GOARCH=arm64 go build ./...` ✓
- `go test ./... -short` ✓ · `go test -race -short ./pipeline/ ./agent/` ✓
- `dippin validate examples/build_product.dip` ✓
- `dippin doctor` on the three core pipelines — A/95, A/90, A/100 ✓
- `dippin simulate -all-paths examples/build_product.dip` — 100 paths, all terminate ✓

## Out of scope

#304 (cost/no-progress detector), #310, #301, #305/#306, #313. No new handler/node type (reuses `wait.human`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Operator-decision node + warm `continue +N` for steady-progress turn breaches**
+  (closes #318, completes the #303 PR2 / epic #308 Phase 1 turn-limit track). When a
+  native-backend turn-limit breach classifies to `operator_decision` (PR1 #317: steady
+  progress, verify ran but not green, no loop), the operator now gets a real choice
+  instead of a bare escalation. In `build_product.dip`, `Implement` routes
+  `when ctx.turn_breach_class = operator_decision` to a new `OperatorDecision`
+  `wait.human` (freeform) gate — **no new handler** — with four labeled edges:
+  `stop` (escalate, preserving work), `abandon` (end the run), `commit_advance`
+  (persist the tree via `CommitIfDirty` and advance), and `continue` (warm retry).
+  - **Deterministically safe unattended default:** the gate pins `default: stop` and
+    lists `stop`/`abandon` *before* `continue`, so `--auto-approve` / `--webhook-url`
+    (and the timeout path) resolve to `stop` — an unattended run never silently
+    advances unverified work. (`--autopilot lax`/`mentor` bias to forward progress and
+    are not deterministically safe; the gate prompt frames `continue` as the risky
+    option.) A pathological breach still narrowly falls through to the existing
+    `EscalateMilestone` catch-all.
+  - **Warm `continue +N`:** a new node-scoped, disk-driven `MaxTurns` override
+    (`.tracker/turn_overrides/<nodeID>`, consulted in `codergen.buildConfig`) lets the
+    capped `ContinueWithMoreTurns` tool node re-enter `Implement` with a bumped turn
+    budget while `PriorEpisodeSummaries` carry across warm. The cap is a per-loop disk
+    counter (not the global engine `RestartCount`); past the cap it escalates instead
+    of looping.
+
 - **Graph-level `on_failure` default failure route** (closes #309, refs epic #308,
   pairs with the #295 strict-failure catch-all). dippin's workflow-level
   `defaults { on_failure: <NodeID> }` now reaches the engine: the adapter
--- a/examples/build_product.dip
+++ b/examples/build_product.dip
@@ -727,6 +727,18 @@ workflow BuildProduct
       BUMP=40
       OVR_DIR=".tracker/turn_overrides"
       ATTEMPT_FILE="$OVR_DIR/continue_attempts"
+      # Keep these tracker-internal control files OUT of the product repo. A later
+      # CommitIfDirty runs `git add -A`, which would otherwise commit the counter +
+      # override — polluting the user's repo AND leaving a stale Implement override
+      # that a future run would read (via codergen.buildConfig) before any operator
+      # decision. Ignore them via the LOCAL, untracked .git/info/exclude so we never
+      # touch the user's tracked .gitignore (idempotent; safe outside a git repo).
+      GITDIR=$(git rev-parse --git-dir 2>/dev/null || true)
+      if [ -n "$GITDIR" ]; then
+        mkdir -p "$GITDIR/info"
+        grep -qxF "$OVR_DIR/" "$GITDIR/info/exclude" 2>/dev/null \
+          || echo "$OVR_DIR/" >> "$GITDIR/info/exclude"
+      fi
       mkdir -p "$OVR_DIR"
       ATTEMPTS=0
       if [ -f "$ATTEMPT_FILE" ]; then
--- a/examples/build_product.dip
+++ b/examples/build_product.dip
@@ -703,6 +703,47 @@ workflow BuildProduct
       fi
       printf 'commit-if-dirty-done'
 
+  tool ContinueWithMoreTurns
+    label: "Warm continue: bump turn budget (capped)"
+    timeout: 5s
+    command:
+      set -eu
+      # Issue #318 warm continue+N. The operator picked "continue" at the
+      # OperatorDecision gate: re-enter Implement WARM (it keeps its episode
+      # memory across the restart) but with a larger turn budget so it doesn't
+      # breach at the same wall again.
+      #
+      # Per-loop circuit-breaker: a disk counter — NOT the global engine
+      # RestartCount, which is shared across every loop in the run and would
+      # reset/consume budget with the wrong semantics (#318 hazard 3). Once the
+      # cap is exhausted we route to EscalateMilestone via ctx.outcome = fail
+      # (exit 1) instead of looping forever.
+      #
+      # The bump is delivered through the tracker-owned, node-scoped MaxTurns
+      # override file that codergen.buildConfig consults
+      # (.tracker/turn_overrides/<nodeID>); BASE matches Implement's max_turns.
+      CAP=3
+      BASE=50
+      BUMP=40
+      OVR_DIR=".tracker/turn_overrides"
+      ATTEMPT_FILE="$OVR_DIR/continue_attempts"
+      mkdir -p "$OVR_DIR"
+      ATTEMPTS=0
+      if [ -f "$ATTEMPT_FILE" ]; then
+        ATTEMPTS=$(cat "$ATTEMPT_FILE")
+      fi
+      ATTEMPTS=$((ATTEMPTS + 1))
+      echo "$ATTEMPTS" > "$ATTEMPT_FILE"
+      if [ "$ATTEMPTS" -gt "$CAP" ]; then
+        echo "continue cap ($CAP) exhausted after $ATTEMPTS attempt(s) — escalating"
+        printf 'continue-cap-exhausted'
+        exit 1
+      fi
+      NEWMAX=$((BASE + ATTEMPTS * BUMP))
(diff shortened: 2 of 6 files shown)


clintecker:
Thanks — all three addressed in 5819756:

**P1 — override files committed by `CommitIfDirty` (`git add -A`).** Real and serious: it both polluted the product repo and left a stale `Implement` override that a *future* run would read via `codergen.buildConfig` before any operator decision. Fixed by having `ContinueWithMoreTurns` add `.tracker/turn_overrides/` to the **local, untracked `.git/info/exclude`** before writing (idempotent; never touches the user's tracked `.gitignore`). Verified in a scratch repo: `git add -A` no longer stages the counter/override, and `git status --porcelain` ignores them.

**Path traversal via `nodeID`.** `readMaxTurnsOverride` now constrains `nodeID` to a bare filename (`safeNodeFilename` — rejects separators / `.` / `..`) so it can't traverse out of the override dir, and only honors a **regular file** (`os.Lstat`) so a planted symlink isn't followed. Added RED→GREEN tests for both (a genuinely-escaping `../../../secret` case and a symlink case). Validation/parse were factored into small helpers to keep every new function within the ≤8 complexity gate.

**Misleading test comment.** Correct — rewrote it: auto-approve resolves to `stop` via **edge ordering** (the freeform `labels[0]` fallback), not because a freeform `default` attr is honored (dippin's `default:` maps to `default_choice`, which freeform mode does not read), and `--webhook` is a separate interviewer not exercised by that test. The `.dip` and the pipeline-package test were already corrected to reflect this in the original commit.

Re-verified: `dippin validate` clean, `doctor` A/90 (baseline held), `simulate -all-paths` 100 paths all terminate, `go test ./... -short` + `-race ./pipeline/ ./agent/` green.

clintecker:
These two were re-emitted by the re-review of 5819756 (Copilot re-scans the full diff each commit), but both are already resolved in that commit:

- **`codergen.go` path traversal:** `readMaxTurnsOverride` now guards the `filepath.Join` with `safeNodeFilename` (rejects any `nodeID` that isn't a bare filename — separators, `.`, `..`, absolute paths) and only honors a regular file via `os.Lstat` (so a planted symlink isn't followed). Covered by `TestReadMaxTurnsOverride_RejectsPathTraversal` (a genuinely-escaping `../../../secret` case) and `TestReadMaxTurnsOverride_SkipsNonRegularFile`.
- **Test comment:** `TestOperatorDecision_UnattendedResolvesToStop`'s comment was rewritten to state the safety comes from EDGE ORDERING (the freeform `labels[0]` fallback), not a freeform `default` attr, and that `--webhook` is a separate path.

No further change needed for these.

clintecker:
Round-3 addressed in bc01b21:

- **`resolveMaxTurns` unconditional honor / cross-workflow reuse:** a warm continue is now **raise-only and bounded** — an override that doesn't exceed the node's base `max_turns` (stale, or left by a different workflow reusing a node ID like `Implement`) is ignored, and one past `maxTurnsOverrideCeiling` (1000) is ignored so a corrupted file can't inflate runtime/cost (BudgetGuard remains the global cost backstop). I declined full per-run path namespacing: working-dir scoping + the Setup/MarkMilestoneDone resets + `.git/info/exclude` already bound the staleness, and threading the runID into the shared Go-constant path is heavier for little additional safety here. New RED→GREEN tests cover the decrease and oversize cases.

- **`continue_attempts` integer assumption / stale state:** `ContinueWithMoreTurns` now resets a non-numeric/corrupted counter to 0 (so the `set -e` arithmetic can't crash), and `Setup` clears `.tracker/turn_overrides` at the start of a **fresh** run. Setup is skipped on checkpoint resume, so a deliberately-resumed run keeps its counter while a fresh run always gets its full continue allowance.

- **Test helper hard-coded path:** `writeTurnOverride` and the symlink test now use the in-package `maxTurnsOverrideSubdir` constant, so they track the override location if it moves.

Re-verified: `dippin validate` clean, `doctor` A/90, `simulate -all-paths` 100 paths terminate, all new functions ≤8 complexity, `go test ./... -short` green.

Issue #422, opened 2026-06-25
engine: sleep-aware wall-clock budgets and sub-node checkpointing for graceful suspend/resume


clintecker:
## Context
When the operator closed their laptop mid-run (the ~17h gap in run `b65fb64ac099` — NOT a tracker fault), two real engine gaps surfaced as a fair enhancement ask.

## Problem
1. The wall-clock budgets COUNT SLEEP TIME. BudgetGuard enforces MaxWallTime (#67) and stall_timeout (#310) via `time.Since` on stored timestamps (CHANGELOG confirms a per-node progress clock reset only on node completion), so a suspended laptop burns the wall-clock/stall budget while doing zero work and can spuriously trip OutcomeBudgetExceeded on resume.
2. Resume granularity is the whole completed node: checkpoint/resume only captures terminal-node boundaries, so a multi-turn agent node (Implement runs up to 50 turns, build_product.dip:904) interrupted mid-node loses that node's entire in-progress work on resume.

## Evidence
run `b65fb64ac099` had a ~17h laptop-closed gap. BudgetGuard MaxWallTime (#67 §0.20.0), stall_timeout per-node progress clock via stored timestamps (#310 §0.39.0). Checkpoint/resume at terminal-node granularity (ResumeRunID §0.18.x; checkpoint tags §0.18.0, 'skips run-started commit on resume'). Implement `max_turns: 50` at build_product.dip:904.

## Proposal
(A) Sleep-aware budgets — detect wall-clock discontinuity (monotonic vs wall-clock divergence, or a sub-threshold heartbeat gap) and EXCLUDE detected sleep from MaxWallTime and stall_timeout accounting. Add Pause()/Resume() on BudgetGuard that subtracts the suspended interval. (B) Sub-node checkpoint hook — let a long-running agent node emit an intra-node checkpoint at turn boundaries (persist conversation/episode state + working-tree SHA) so resume re-enters mid-node rather than discarding the node's progress. Both opt-in via graph attrs to preserve current strict-budget semantics for callers who want sleep counted.

## Acceptance criteria
1. A simulated suspend (clock jump while the engine is idle) does not advance MaxWallTime or stall_timeout accounting (inject a clock with a gap; assert the budget is not tripped).
2. BudgetGuard exposes Pause/Resume (or equivalent) subtracting the suspended span.
3. An agent node interrupted between turns resumes from its last turn checkpoint with episode state intact.
4. Default behavior unchanged when the new attrs are absent.

## Genericity
[...]

PR #425, opened 2026-06-25
engine: opt-in content-hash node-output memoization (#421)

### LINKED ISSUE #421: engine: content-hash node-output memoization to skip already-green work across loop restarts
## Context
run `b65fb64ac099` fired 11 loop-restarts. Every restart edge re-executes already-green downstream nodes via the clearDownstream BFS, and the engine has NO node-output memoization — the only caching is per-session `cache_tools`/CacheToolResults and Anthropic prompt-cache token accounting, neither of which skips re-running a node whose inputs are unchanged.

## Problem
Work that survived a restart (a green milestone, a completed SpecLint/ReadSpec) is recomputed, wasting tokens and active wall-clock on a run that already cost $27.13 / 3.42M tokens. clearDownstream re-executes the cleared set on each restart.

## Evidence
11 loop-restarts measured in run `b65fb64ac099`. clearDownstream BFS re-executes the cleared set (CHANGELOG §0.40.0). Only per-session cache_tools (§0.29.0) and prompt-cache accounting (#185) exist — no node-output memo. Restart edges (`restart:true`) at build_product.dip:2545 (MarkMilestoneDone -> PickNextMilestone), :2574-2576 (FixMilestone -> TestMilestone), and the other loop-back edges in the edge block (build_product.dip:2603-2651).

## Proposal
Add an opt-in content-hash node-output cache keyed on (node id + hash of resolved inputs: prompt/command after interpolation, upstream ctx values it reads, relevant file/tree SHA). On re-entry to a node whose key matches a prior successful execution within the same run, replay the stored outcome + ctx delta instead of re-running the handler. Gate behind a node/graph attr (`memoize: true`) so authors opt in only where outputs are pure w.r.t. the hashed inputs; tool nodes with working-tree side effects key on the tree SHA. Must compose with checkpoint resume (persist the memo table in checkpoint.json).

## Acceptance criteria
1. A `memoize:true` node re-entered via a restart edge with identical hashed inputs replays its prior outcome without invoking the handler (handler call count == 1 across 2 entries).
2. Any change to a hashed input (ctx value, interpolated prompt, or keyed tree SHA) invalidates the entry and re-runs.
3. The memo table survives checkpoint resume.
4. Off by default; no behavior change for existing .dip files lacking the attr.

## Genericity
Engine-only and product-agnostic: keys on generic node inputs and tree SHAs; no knowledge of milestones, languages, or build_product structure.

## Attribution
engine.

clintecker:
## Summary

Closes #421. Adds an **opt-in, content-hash node-output memo cache** so a loop-restart that re-enters an already-green node with identical inputs **replays** the stored outcome instead of re-invoking the handler. Run `b65fb64ac099` fired 11 loop-restarts, each re-executing the cleared set via `clearDownstream`; there was no node-level memoization (the only caches — the per-session tool cache and Anthropic prompt-cache token accounting — never skip a node).

## How it works

- **Opt-in** via `params: { memoize: true }` on an agent node (read as `node.Attrs["memoize"]` through the generic params pass-through — **no dippin-lang field**, verified `dippin doctor` Grade A with the attr). Off by default: a `.dip` without the attr builds no memo table and runs the handler on every entry (AC4).
- **Memo key** = SHA-256 (length-prefixed, anti-collision) over: a recipe-version tag, node id, handler, the sorted post-interpolation `node.Attrs` (so the resolved prompt embeds every templated `ctx.*`), a filtered bare-namespace context snapshot, and (for agent nodes only) nothing more. A `writable_paths` node is an **unconditional** hard miss — never memoized, so no working-tree fingerprint enters the key (see note below). Documented verbatim in `computeMemoKey`.
- **Replay** before handler dispatch on a key hit; **store** only on `OutcomeSuccess` (failures never replay). The replay reuses the identical post-handler tail (`finishNode`), so edge routing and strict-failure semantics are untouched.
- **Persisted** in `checkpoint.json` (additive, `omitempty`), so the table survives resume (AC3); `clearDownstream` never clears it.
- **Conservative by construction**: a `writable_paths` node is always a hard cache **miss** → the handler re-runs (never replayed); more generally, any uncertainty means a miss. Never replay on an uncertain key.

## Scope: agent-node-only (documented narrowing)

`dippin v0.43.0` rejects `params:` on **tool** nodes and rejects `defaults: memoize` — so there is no `.dip` surface to opt a tool node in without a dippin-lang field (forbidden by the brief). #421 therefore ships **agent-node memoize only**; an agent that declares `writable_paths` is an unconditional hard miss (never memoized — the artifact-repo fingerprint can't validate the agent's separate `working_dir`, so side-effecting nodes always re-run; `TreeFingerprint` is retained as a building block for a future `working_dir` fingerprint but is unused by the memo key today). Tool-node opt-in is a follow-up (needs a dippin-lang `ToolConfig.Params` field). Called out in the CHANGELOG.

## Adversarial review: two blockers found and fixed

A 3-lens adversarial panel caught two **stale-replay** holes (the exact failure mode #421 must prevent); both were fixed with TDD and re-verified by an independent reviewer:

1. **Injected prompt inputs masked as self-outputs.** `InjectPipelineContext` appends bare `last_response`/`human_response` to every agent prompt; the original self-output exclusion dropped them from the key, so a `build→review→fail→build` loop replayed a stale build after the critique changed. **Fix:** those two injection keys are now hashed unconditionally (other bare self-outputs keep the exclusion). New test `TestMemoInvalidatesOnChangedInjectedLastResponse`.
2. **Tree-fingerprint was dead code in the shipping config (later superseded).** `WithGitArtifacts` has no production callers, so `s.gitRepo` is nil by default and a `writable_paths` node would have replayed without tree-drift detection. A follow-up review round made `writable_paths` an **unconditional** hard miss regardless of repo availability — the fingerprint is taken from the artifact repo (`<artifactDir>/<runID>`), a different directory than the agent's `working_dir`, so it could never validate a side-effecting node's inputs. Net shipped behavior: `writable_paths` nodes are never memoized.

**Known limitation (pre-existing, non-blocking):** under non-`FidelityFull` prompt fidelity, the compacted-context summary could in principle inject a non-`last_response` bare key that the self-output exclusion still masks. This is a narrow, pre-existing class (the demonstrated real-world loop uses `last_response`, now always hashed) and is not a regression from this change; noted for awareness.

## Verification

- `go build ./...` ✅ · `go test ./... -short` ✅ · `go test -race ./pipeline/ -run 'Memo|Restart|Checkpoint'` ✅
- Tests cover all four ACs (replay-skips-handler with call-count==1; ctx/prompt/tree invalidation; checkpoint round-trip; off-by-default) plus the two blocker-regression tests.
- `dippin doctor examples/build_product.dip` — **Grade A**; tolerates the `memoize` attr with no new lint error.
- Pre-commit hooks (gofmt, vet, build, full suite, race, coverage 85%, complexity ≤8, dippin lint) all pass (a minimal `isMemoizableContextKey` extraction kept `memoizableContext` under the complexity gate).

Closes #421

---
<a href="https://app.blacksmith.sh/2389-research/codesmith/tracker/pr/425"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785006503&installation_id=131176874&pr_number=425&repository=2389-research%2Ftracker&return_to=https%3A%2F%2Fgithub.com%2F2389-research%2Ftracker%2Fpull%2F425&signature=c028abd74bfedb31ce89858aa1397866361474bf9a2390a717f543858f3b0efa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added opt-in SHA-256 content-hash memoization for node outputs (`memoize: true`), replaying prior successful outcomes on identical re-entry after restarts.
  * Memoized replays are persisted in checkpoints and emit a new `node_memo_replayed` pipeline event.
* **Bug Fixes**
  * Hardened replay correctness by invalidating on changed resolved inputs/prompts and ensuring replay follows the same routing and terminal behavior as the original run.
* **Tests**
  * Added coverage for replay-on-restart, memo invalidation rules, checkpoint persistence/round-trips, deterministic memo keying, deep-copy safety, and tree fingerprinting behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and the node re-runs live rather than replaying a stored failure into routing.
   A node declaring `writable_paths` is an **unconditional** hard
   miss — it always re-runs and is never replayed, with no warning (a by-design
-  policy skip, not a key-computation failure). Such a node mutates and reads the
+  policy skip, not a key-computation failure). The miss is keyed on attr
+  **presence**, not a non-empty value, mirroring `AgentConfig.WritablePathsSet`
+  and the fs-jail gate — so the adapter's `writable_paths: ""` bypass-defense
+  sentinel still refuses memoization (#425 review). Such a node mutates and reads the
   agent's session `working_dir`, but tracker can only fingerprint the ARTIFACT
   repo (`<artifactDir>/<runID>`), a different directory — so any content
   fingerprint would prove nothing about the tree the agent actually read/wrote
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   file edits that the prior staged-blob listing missed); the persisted
   `MemoEntry` now carries the `PreferredLabel`/`SuggestedNextNodes` routing hints
   so a replayed node follows the same edge the original did; and replayed trace
-  entries are stamped with a real timestamp.
+  entries are stamped with a real timestamp. Second review round: the worktree
+  fingerprint now fails closed on an unexpected `read-tree HEAD` error (only a
+  genuinely missing HEAD proceeds with an empty index — otherwise a tracked
+  deletion could yield a stale fingerprint / false cache hit); and the replay
+  Outcome deep-copies its `ContextUpdates`/`SuggestedNextNodes` so downstream
+  mutation can never corrupt the persisted memo record.
 
 ## [0.40.2] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   genuinely missing HEAD proceeds with an empty index — otherwise a tracked
   deletion could yield a stale fingerprint / false cache hit); and the replay
   Outcome deep-copies its `ContextUpdates`/`SuggestedNextNodes` so downstream
-  mutation can never corrupt the persisted memo record.
+  mutation can never corrupt the persisted memo record; and the replay path now
+  threads the run's `ctx` (instead of `context.Background()`) into `finishNode`
+  so cancellation/deadlines still apply and a non-success memo record from a
+  corrupted/hand-edited checkpoint routes through the same ctx-aware tail.
 
 ## [0.40.2] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Opt-in content-hash node-output memoization (#421).** Annotate an agent
+  node with `params: { memoize: true }` and a loop-restart that re-enters the
+  already-green node with identical resolved inputs now REPLAYS the stored
+  successful outcome instead of re-running the handler. The memo key is a
+  SHA-256 over the node's resolved attrs (post-interpolation prompt/command)
+  and the bare-namespace context it reads — including the `last_response` and
+  `human_response` prompt inputs injected into every agent prompt, which are
+  hashed unconditionally so an intervening node's new critique invalidates the
+  replay. Memoization is **agent-node-only**. A node declaring `writable_paths`
+  has working-tree side effects and **requires a working-tree fingerprint to
+  replay**: without a live git repo (or on a fingerprint error) it is a hard
+  cache miss and re-runs, never replaying on an unproven tree. Any input change
+  yields a different key and re-runs the node. Only successful outcomes are
+  memoized (failures never replay), the memo table is persisted in
+  `checkpoint.json` so replay survives resume. Off by default: a `.dip` file
+  lacking the attr has byte-identical behavior and writes no `memo_entries`.
+  Emits `node_memo_replayed` on a replay.
+
 ## [0.40.2] - 2026-06-24
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   memoized (failures never replay), the memo table is persisted in
   `checkpoint.json` so replay survives resume. Off by default: a `.dip` file
   lacking the attr has byte-identical behavior and writes no `memo_entries`.
-  Emits `node_memo_replayed` on a replay.
+  Emits `node_memo_replayed` on a replay. Review hardening: the memo key now
+  re-includes a bare key the node self-aliased on a prior pass once an
+  intervening node OVERWRITES it (so replay can't serve stale output against a
+  changed input); the `writable_paths` tree fingerprint now hashes actual
+  worktree CONTENT via a throwaway-index `write-tree` (catching unstaged/untracked
+  file edits that the prior staged-blob listing missed); the persisted
+  `MemoEntry` now carries the `PreferredLabel`/`SuggestedNextNodes` routing hints
+  so a replayed node follows the same edge the original did; and replayed trace
+  entries are stamped with a real timestamp.
 
 ## [0.40.2] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `human_response` prompt inputs injected into every agent prompt, which are
   hashed unconditionally so an intervening node's new critique invalidates the
   replay. Memoization is **agent-node-only**. A node declaring `writable_paths`
-  has working-tree side effects and **requires a working-tree fingerprint to
-  replay**: without a live git repo (or on a fingerprint error) it is a hard
-  cache miss and re-runs, never replaying on an unproven tree. Any input change
+  has working-tree side effects and is **never memoized** — an unconditional hard
+  cache miss, so it always re-runs (see the third review-round note below for why).
+  Any input change
   yields a different key and re-runs the node. Only successful outcomes are
   memoized (failures never replay), the memo table is persisted in
   `checkpoint.json` so replay survives resume. Off by default: a `.dip` file
@@ -41,7 +41,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mutation can never corrupt the persisted memo record; and the replay path now
   threads the run's `ctx` (instead of `context.Background()`) into `finishNode`
   so cancellation/deadlines still apply and a non-success memo record from a
-  corrupted/hand-edited checkpoint routes through the same ctx-aware tail.
+  corrupted/hand-edited checkpoint routes through the same ctx-aware tail. Third
+  review round: a `writable_paths` node is now an **unconditional** hard miss
+  rather than replaying on an artifact-repo tree fingerprint. The fingerprint was
+  taken from `s.gitRepo` — the artifact repo at `<artifactDir>/<runID>`, a
+  different directory than the agent's session `working_dir` where `writable_paths`
+  actually takes effect — so it proved nothing about the tree the agent read/wrote
+  and could have replayed against a changed working tree. Until tracker can
+  fingerprint the agent's real `working_dir`, side-effecting nodes are simply not
+  memoizable (the `TreeFingerprint` helper is retained as the building block for
+  that follow-up).
 
 ## [0.40.2] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,9 +34,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timestamp; the replay Outcome deep-copies its `ContextUpdates`/`SuggestedNextNodes`
   so downstream mutation can never corrupt the persisted memo record; and the
   replay path threads the run's `ctx` (instead of `context.Background()`) into
-  `finishNode` so cancellation/deadlines still apply and a non-success memo
-  record from a corrupted/hand-edited checkpoint routes through the same
-  ctx-aware tail. A node declaring `writable_paths` is an **unconditional** hard
+  `finishNode` so cancellation/deadlines still apply; and the replay lookup now
+  enforces the "failures never replay" contract on the READ side too — a
+  non-success memo record (only reachable via a corrupted/hand-edited checkpoint,
+  since the store site is gated on `OutcomeSuccess`) is treated as a cache miss
+  and the node re-runs live rather than replaying a stored failure into routing.
+  A node declaring `writable_paths` is an **unconditional** hard
   miss — it always re-runs and is never replayed, with no warning (a by-design
   policy skip, not a key-computation failure). Such a node mutates and reads the
   agent's session `working_dir`, but tracker can only fingerprint the ARTIFACT
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   hashed unconditionally so an intervening node's new critique invalidates the
   replay. Memoization is **agent-node-only**. A node declaring `writable_paths`
   has working-tree side effects and is **never memoized** — an unconditional hard
-  cache miss, so it always re-runs (see the third review-round note below for why).
+  cache miss, so it always re-runs (see the `writable_paths` note below for why).
   Any input change
   yields a different key and re-runs the node. Only successful outcomes are
   memoized (failures never replay), the memo table is persisted in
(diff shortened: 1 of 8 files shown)


clintecker:
Pushed `eb52485` addressing all four review findings:

- **Routing hints (CodeRabbit, major):** `MemoEntry` now persists `PreferredLabel` + `SuggestedNextNodes`; `replayMemoizedNode` reconstructs the `Outcome` with them, so a memoized node replays onto the **same edge** the original execution selected.
- **Worktree content fingerprint (Codex P2 / CodeRabbit, major):** `TreeFingerprint` now hashes actual worktree **content** via a throwaway `GIT_INDEX_FILE` (`add -A` + `write-tree`) instead of the staged-blob `ls-files -s` listing — unstaged-tracked and untracked content edits now flip the key. Side-effect-free (real index untouched; dangling objects gc'd, same posture as `git stash create`).
- **Self-output exclusion narrowed (Codex P1):** a bare key the node self-aliased on a prior pass is skipped **only while unchanged**; an intervening overwrite (bare ≠ scoped) makes it a genuine changed input that is hashed again, preventing stale replay.
- **Replay timestamp (CodeRabbit, minor):** the synthesized replay `TraceEntry` now stamps `time.Now()`.

New tests: `TestMemoEntryPersistsRoutingHints`, `TestTreeFingerprintDetectsUnstagedTrackedContentChange`, and overwritten-self-output cases in `TestComputeMemoKey`. Default-off behavior unchanged. Verified: build, `go test ./... -short`, race on `pipeline`, `dippin doctor` Grade A, all pre-commit gates (coverage 85.1%, complexity ≤8).

clintecker:
Addressed the 06-29 review round:

**Copilot — `replayMemoizedNode` dropped the run context:** It called `finishNode(context.Background(), ...)`, discarding the run's cancellation/deadline. Threaded the run's `ctx` through `maybeReplayMemoized` → `replayMemoizedNode` → `finishNode`, so cancellation still applies and a non-success memo record (e.g. a corrupted or hand-edited checkpoint) routes through the same ctx-aware tail as a live execution.

**Copilot — deep-copy `ContextUpdates`/`SuggestedNextNodes`:** Already addressed in `85503ac` (prior round) via the `memoOutcome` helper, which deep-copies the reference-typed fields so the replay `Outcome` (and the `s.lastOutcome` copy) can never mutate the persisted memo record. Verified `TestMemoOutcomeIsolatesRecordFromMutation` still passes.

`go build`, memoization tests, and the full pre-commit gates (race, coverage, complexity, dippin lint) are green.

clintecker:
## Addressed latest review feedback (commit 155cd72)

**Copilot @ `pipeline/memoize.go:92` — writable_paths tree fingerprint comes from the artifact repo, not the agent's working_dir → stale-replay risk.**

You're right, and it's a soundness bug. `s.gitRepo` is the **artifact** repo rooted at `<artifactDir>/<runID>` (`initGitRepo` in engine_run.go), whose `TreeFingerprint()` runs `git status` / `write-tree` in *that* directory. But `writable_paths` constrains the agent's session `working_dir` (backend_native.go) — a different directory. So the fingerprint proved nothing about the tree the agent actually read/wrote: a node could replay while its real working tree had changed, as long as the artifact dir was unchanged. That's exactly the stale-replay class this feature is meant to prevent.

Per the feature's own stated rule — *"never replay on an uncertain key"* — an artifact-repo fingerprint **is** an uncertain key for the working tree. Fix:

- A `writable_paths` node is now an **unconditional hard miss** (`ok=false` → handler re-runs), regardless of whether a git repo is present. The misdirected artifact-fingerprint path is removed.
- **No production behavior change:** `WithGitArtifacts` has no production callers, so `s.gitRepo` was already nil (hard miss) in shipping configs. This only removes the unsound path a future git-artifacts wiring would have exercised.
- `TreeFingerprint` / `worktreeContentTree` are **retained** (still unit-tested via `TestTreeFingerprintDetectsUnstagedTrackedContentChange`) as the building block for the proper follow-up: fingerprinting the agent's real `working_dir`.

Tests: rewrote the AC2(iii) sensitivity test as `TestMemoKeyWritablePathsAlwaysMissesWithLiveRepo` (asserts hard miss even with a live artifact repo), renamed the nil-repo test to `TestMemoKeyWritablePathsMissesWithoutRepo`, and dropped the now-unreachable `TestMemoKeyTreeFingerprintErrorIsMiss`. CHANGELOG updated with a third review-round note.

Gates: build, test -short, race, `pipeline` coverage 85.1%, complexity ≤ 8, dippin lint — all green.

clintecker:
## Merged `origin/main` (dbe6780) — conflicts resolved

No outstanding review feedback on this PR; this was a conflict-only sync after #426 (sleep-aware budgets) merged to main.

- **`CHANGELOG.md` `[Unreleased]`** — kept both this PR's "Opt-in content-hash node-output memoization (#421)" and #426's "Sleep-aware budgets (#422, part A)" entries under `### Added`, plus #426's `### Notes` block.
- **`pipeline/engine.go`** — both branches independently refactored the terminal-success helper out of `Run` (this PR → `buildSuccessResult`, #426 → split `done:` + `successResult`). The textual auto-merge left a dangling marker block and a duplicate method. Resolved to this PR's self-contained `buildSuccessResult` (finalizes trace, emits `EventPipelineCompleted`, builds the override-aware terminal result in one function); dropped #426's now-redundant `successResult`. The override-aware status logic is byte-identical between the two.

All pre-commit gates green (build, vet, test -short, race, coverage, complexity, dippin lint).

clintecker:
Addressed the fresh review feedback (commit 7632ca2):

**1. Misleading "skipped memoization" warning on a hard miss** (`pipeline/engine.go`, `maybeReplayMemoized`)
`computeMemoKey` has no runtime-error path — both `ok=false` returns are deterministic policy (`!memoizeEnabled`, pre-checked by the caller; and a `writable_paths` node, an unconditional hard miss). So emitting an `EventWarning` ("could not compute memo key…") implied a swallowed computation failure that can't occur. Removed the warning entirely: a non-memoizable node now just runs the handler live, which is the by-design behavior. (Not a CLAUDE.md "silent swallow" — there is no error to surface.)

**2. `pendingMemoKey` struct comment** (`pipeline/engine_run.go`)
Reworded from "tree fingerprint is computed once per entry" to describe the actual memo key, and noted it is empty when the node didn't opt in or isn't memoizable (`writable_paths` → unconditional hard miss).

**3. CHANGELOG reads as if memoization uses a `writable_paths` fingerprint today** (`CHANGELOG.md:33`, Copilot)
The entry carried the intra-PR review narrative: round-1/2 described a `writable_paths` tree fingerprint ("hashes worktree CONTENT via a throwaway-index `write-tree`", "fails closed on `read-tree HEAD` error"), which round-3 then superseded by making `writable_paths` an unconditional hard miss that computes no fingerprint. Collapsed the superseded clauses into the shipped behavior: `writable_paths` is never memoized (the artifact repo `<artifactDir>/<runID>` is a different directory than the agent's `working_dir`, so a fingerprint would prove nothing); `TreeFingerprint` is retained as a building block but is not used in the memo key today. Also fixed the now-stale "see the third review-round note below" cross-reference.

Gates green (build, 19-pkg tests, race, pipeline coverage 85.2%, complexity ≤8, dippin lint A).

clintecker:
## Latest Copilot round addressed (commit 63bbef7 + PR description)

**1. `maybeReplayMemoized` replayed any hit regardless of stored `Status`** (`engine.go:386`). `PutMemo` is gated on `OutcomeSuccess` at the call site, so a non-success record can only arise from a corrupted/hand-edited checkpoint — but the read side replayed it anyway, contradicting the "failures never replay" contract and risking a stored failure being routed. Added a guard: a non-success `MemoEntry` is now a cache **miss**, so the node re-runs live. New regression test `TestMemoNonSuccessRecordIsNotReplayed` (injects a fail record under the computed key, asserts no replay + the pending key is still recorded for the live re-run's store).

**2. `TreeFingerprint` doc claimed it powers memoization** (`git_artifacts.go:244`). It does not: a `writable_paths` node is an unconditional hard miss, and the fingerprint is taken from the artifact repo (`<artifactDir>/<runID>`) — a different directory than the agent's session `working_dir` where side effects land — so it can't validate a side-effecting node's inputs. Reworded: `TreeFingerprint` is a retained building block for a future `working_dir` fingerprint, unused by the memo key today.

**3. PR description reconciled** (`memoize.go:93`). The body described the memo key as including a `writable_paths` working-tree fingerprint. Updated the memo-key definition, the "conservative by construction" item, the agent-only-scope paragraph, and the review-round note so they all describe the shipped behavior: `writable_paths` nodes are an unconditional hard miss (never memoized).

**Also folded in this round:** CHANGELOG documents the read-side success guard.

All pre-commit gates green (build, vet, 19-pkg tests, race, pipeline coverage 85.2%, complexity ≤8, dippin lint A).

clintecker:
Addressed the latest Copilot finding (commit 84cae56):

**`computeMemoKey` hashed before the `writable_paths` short-circuit** (`memoize.go:93`). The full attrs + bare-context snapshot + SHA-256 ran *before* the `writable_paths` unconditional-hard-miss check, wasting that work for a side-effecting node mistakenly marked `memoize:true`. Moved the check to the top (right after the `memoizeEnabled` gate), before any hashing. Behavior is identical (same `ok=false` result) — purely avoids the dead hash work.

All pre-commit gates green (build, vet, 19-pkg tests, race, pipeline coverage, complexity ≤8, dippin lint A).

clintecker:
Addressed the new Copilot finding (commit 71a8776):

**`pipeline/memoize.go` — `writable_paths` hard miss was keyed on non-emptiness, not presence.** The check was `strings.TrimSpace(attr) != ""`, which lets a present-but-empty `writable_paths: ""` slip through and be memoized. But the adapter deliberately emits `writable_paths: ""` as a bypass-defense sentinel (`dippin_adapter.go:301-302` — an empty glob set jails the node to *nothing* rather than silently dropping the signal), and both `AgentConfig.WritablePathsSet` (`node_config.go:287`) and the fs-jail gate (`codergen_jail.go`) key on attr **presence**. Memoization now aligns: `if _, ok := execNode.Attrs["writable_paths"]; ok { return "", false }`, so a present-but-empty value is still treated as a side-effecting node and never replayed.

Added `TestMemoKeyWritablePathsEmptyStillMisses`. CHANGELOG updated. Full gates green (race, pipeline coverage 85%+, complexity, lint).

clintecker:
## Copilot review (2026-06-30) — addressed in `de56f94`

**`pipeline/memoize.go:21` — `memoizeEnabled` did not enforce the agent-node-only scope.**

Fixed. `memoizeEnabled` previously checked only the `memoize` attr, so a `tool` / `wait.human` / `conditional` node carrying `memoize: true` (reachable via a DOT graph or a programmatic `Graph` — the dippin adapter only surfaces the attr for agent nodes) would have been replayed, skipping a side-effecting or non-deterministic handler. That made the actual scope broader than the documented "agent-node-only" contract.

The guard now requires `execNode.Handler == "codergen"` in `memoizeEnabled` itself, enforcing the scope in code rather than relying on the adapter. Added `TestMemoKeyNonAgentNodeNeverMemoizes` pinning the hard miss for `tool`/`wait.human`/`conditional` handlers.

clintecker:
## Copilot review (2026-06-30) — both findings addressed in `1d5e814`

Both were doc-comment type-reference nits:

- **`pipeline/memoize.go:96`** and **`pipeline/memoize_test.go:555`** referenced `AgentConfig.WritablePathsSet`, but the config type is `AgentNodeConfig` (the `AgentConfig` name is the `Node.AgentConfig(...)` accessor, not the struct that carries `WritablePathsSet`).

Corrected both to `AgentNodeConfig.WritablePathsSet` so the reference resolves cleanly. Comment-only change; full gates pass.

clintecker:
Addressed both findings from this review in 17242a1:

**1. `pipeline/git_artifacts.go` — `worktreeContentTree` read-tree HEAD handling**
The old code treated a `read-tree HEAD` failure as "no HEAD yet" whenever `rev-parse --verify HEAD` *also* failed — which would silently proceed with an empty index even for a **corrupt** (present-but-unreadable) HEAD, contradicting the "fail closed" comment and risking a stale fingerprint (false cache hit). Extracted the seeding into `seedTreeIndexFromHEAD` and reordered to probe HEAD first:
- `rev-parse --verify HEAD` resolves → `read-tree HEAD` **must** succeed; any error is hard (fail closed).
- HEAD doesn't resolve → distinguish truly-unborn (no commits anywhere, via `rev-list -n 1 --all` returning empty → proceed with empty index) from corrupt (any reachable commit → fail closed).

**2. `pipeline/memoize_test.go` — `TestMemoOffByDefault` brittle assertion**
Replaced the `strings.Contains(string(data), "memo_entries")` substring check (which could false-positive on an unrelated field's value) with a parse into `map[string]any` and an explicit key-presence assertion. Dropped the now-unused `strings` import.

Full pre-commit gates pass (gofmt/vet/build, 19-pkg tests, race detector, pipeline coverage 85.4%, complexity ≤8 — extracting the helper kept `worktreeContentTree` under the gocyclo threshold, dippin lint).

Issue #423, opened 2026-06-25
engine: sandbox device-node hygiene + git-artifact durability across long/suspended runs


clintecker:
## Context
On resume after the long suspend in run `b65fb64ac099`, the sandbox's `/dev/null` became unreadable, which broke git and a reviewer CLI, and FinalCommit failed with 'context canceled' while the engine emitted the warning 'cannot preserve uncommitted work (git artifact repository unavailable)'. The never-lose-work guarantee silently degraded to a best-effort warning on the terminal commit path.

## Problem
Several engine/dip helpers depend on a working `/dev/null` (build_product.dip:306 `git hash-object -t tree /dev/null`, :1050 `git diff --no-index --numstat /dev/null`). When the device node or the artifact git repo goes bad mid/post-suspend, the failure surfaces as a deep git error mid-FinalCommit or as a soft EventWarning rather than a loud preflight diagnostic — exactly the never-lose-work degradation observed.

## Evidence
run `b65fb64ac099` resume: `/dev/null` unreadable broke git + reviewer CLI; FinalCommit 'context canceled'; EventWarning 'cannot preserve uncommitted work (git artifact repository unavailable)'. /dev/null dependencies at build_product.dip:306, :1050. Durability scaffolding: WithGitArtifacts (#77 §0.18.0), CommitWIP/commitWIPBeforeRouting wired into fail/exhaust routing (§0.39.1), commit-green-before-escalate (§0.39.1) — none handle artifact-repo unavailability.

## Proposal
(A) Sandbox device-node hygiene — at run start and on resume, verify/repair standard device nodes (`/dev/null` readable+writable) before any git or subprocess handler runs; fail fast with a clear diagnostic (not a deep git error) if they can't be restored, so a broken `/dev/null` is caught at a preflight, not mid-FinalCommit. (B) Artifact-repo health check + reattach — before commitWIPBeforeRouting/FinalCommit relies on the artifact git repo, probe its availability; if it has gone unavailable post-suspend, attempt reattach/reinit to a recoverable ref and surface a hard error rather than degrading the never-lose-work guarantee to a best-effort EventWarning on the terminal commit path.

## Acceptance criteria
1. A run-start/resume preflight detects an unreadable `/dev/null` and emits a specific diagnostic before the first git-dependent node (stubbed device probe).
[...]

PR #426, opened 2026-06-25
feat(budget): sleep-aware wall/stall budgets (#422 part A)

clintecker:
## What & why

Closes #422 (part A).

During the `b65fb64ac099` dogfooding run a closed-laptop suspend (~17h) mid-run burned wall-clock and stall budget while doing **zero work**, spuriously tripping `OutcomeBudgetExceeded` on resume. `BudgetGuard` measured elapsed via `time.Since(started)` and stall via `progressAt` stored as `UnixNano` — which strips Go's monotonic reading — so the stall window absorbed the suspend gap.

## The fix — monotonic-clock accounting (opt-in, default-off)

The only reliable way to distinguish a slept laptop from a slow node is the monotonic clock: Go's monotonic reading (`CLOCK_MONOTONIC`) **freezes during OS suspend** but advances during genuine work.

- `clock` interface now exposes `Now()` (wall) + `Mono() time.Duration` (monotonic elapsed since construction). `realClock` returns `time.Since(start)` for `Mono()`. Tests inject a fake clock whose wall and monotonic readings advance **independently**.
- When `SleepAware` is on, `MaxWallTime` is driven off `Mono()`-since-start and `StallTimeout` off `Mono()`-since-last-progress. A slept laptop is excluded automatically — **no threshold, no heuristic**, and genuine long nodes are still budgeted.
- `Pause()`/`Resume()` subtract an explicit awake-idle span (mutex-guarded, nil-safe) — the opt-in API for awake idle (AC2). Engine wiring around blocking human-gate waits is a documented follow-up.
- **Default-off is byte-identical to today**: with the attr absent, wall uses `now.Sub(started)` and stall uses `time.Unix(0, progressAt)` exactly as before.

Opt-in is via `Config.Budget.SleepAware` + the `--sleep-aware-budget` CLI flag. dippin v0.43.0 rejects a `.dip`-level attr (typed `WorkflowDefaults`), so no `.dip` surface was added.

### Why no raw inter-`Check` threshold

[...]

PR #428, opened 2026-06-25
feat(engine): sandbox device-node hygiene + git-artifact durability (#423)

clintecker:
## Summary

Closes **#423** (PR-D of the b65fb64 review-fidelity & engine-resilience workstream). Hardens two resilience gaps surfaced by the `build_product` dogfood run `b65fb64ac099`, both opt-in/no-op when the relevant conditions are healthy.

### 1. Sandbox device-node hygiene preflight
Before any git or subprocess work, `applyGitPreflight` runs `checkDeviceNodes` to verify standard device nodes are usable (`/dev/null` must be a readable+writable character device). A suspended/restored sandbox can silently corrupt `/dev/null` (becomes unreadable, or a regular file masquerading as the device), which otherwise produces baffling downstream failures. POSIX probe is build-tagged (`device_probe_posix.go`); Windows is a no-op (`device_probe_windows.go`). No new behavior when devices are healthy.

### 2. Artifact-repo health probe + reattach, and never-lose-work hardening
`commitWIPBeforeRouting` now probes artifact-repo availability via `ensureHealthy()` (re-Inits + re-verifies a `.git` lost to a sandbox suspend). When the repo is genuinely unrecoverable, **terminal halt paths** emit an `EventStageFailed` diagnostic and set `EngineResult.WorkPreserveFailed` — surfacing silently-lost work instead of swallowing it. `Status` is never masked (the original node failure stays the primary cause).

## Review remediation (this PR's second commit)

Adversarial review caught that the initial implementation hardened only 2 of the 4 terminal never-lose-work sites. `handleRetryExhausted` and `checkStrictFailure` each have a **no-onward-edge terminal dead-stop branch** that was mis-classified as mid-routing — both discarded the preserve error and never set `WorkPreserveFailed`. Fixed: the preserve error is captured once, still discarded on the genuine mid-routing branch (`fallback_retry_target` / `strictFailureFallback`), and hard-escalated on the terminal sub-branch via `escalateWorkPreserve`. Added `TestRetryExhaustedNoFallback_TerminalHardFail` and `TestStrictFailureNoFallback_TerminalHardFail`.

## Verification
- `go build ./...` ✓
- `go test ./... -short` ✓
- race on `pipeline` + `cmd/tracker` ✓
- `dippin doctor` on the three core pipelines: **Grade A (95/100)** ✓
- pre-commit gates (coverage 85.2%, complexity ≤8, lint, format) ✓

## Residual (unchanged)
Device probe covers `/dev/null` only; network egress and read-side exfil remain out of scope per the #272 jail threat model.

---
<a href="https://app.blacksmith.sh/2389-research/codesmith/tracker/pr/428"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785013365&installation_id=131176874&pr_number=428&repository=2389-research%2Ftracker&return_to=https%3A%2F%2Fgithub.com%2F2389-research%2Ftracker%2Fpull%2F428&signature=865f103b75761af7b6894d9578089962d1cf7ed394d42203e934efc4d1ccf5a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a sandbox device-node hygiene preflight (e.g., `/dev/null`) with actionable remediation, run before any git checks.
  * Added a dedicated terminal “work preserve failed” signal when artifact-repo recovery can’t complete.

* **Bug Fixes**
  * Improved artifact repository health checking with a single reattach/recovery attempt before attempting WIP preservation; detects nested git scenarios to avoid committing to the wrong repo.
  * Preserves the original failure status while escalating only on terminal never-lose-work paths; mid-routing remains warning-only.
  * Kept UI diagnostics consistent for the new terminal escalation case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,13 +35,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the user's enclosing repository.
   On the **terminal** never-lose-work paths (handler-error halt and failing exit
   node) an unrecoverable repo is now surfaced as a HARD signal — a new
-  `EventStageFailed` diagnostic plus an `EngineResult.WorkPreserveFailed` flag —
-  rather than degrading silently to a best-effort `EventWarning`. The original
-  node failure is never masked (`Status` stays `OutcomeFail`). **Mid-routing**
-  call sites (strict-failure pre-decision, retry-exhausted with a
-  `fallback_retry_target`) keep the warning-only behavior so the routing outcome
-  is never changed. No behavior change when devices and the artifact repo are
-  healthy.
+  dedicated `EventWorkPreserveFailed` diagnostic plus an
+  `EngineResult.WorkPreserveFailed` flag — rather than degrading silently to a
+  best-effort `EventWarning`. The signal is a distinct event type (not an extra
+  `EventStageFailed`) precisely so `tracker diagnose` does **not** count it as
+  another per-node execution attempt — emitting it as `stage_failed` would have
+  inflated the failed node's `RetryCount` / `IdenticalRetries` (#428 review).
+  The TUI adapter maps `EventWorkPreserveFailed` to the same `MsgNodeFailed`
+  failure line, so the operator still sees it. The original node failure is never
+  masked (`Status` stays `OutcomeFail`). **Mid-routing** call sites
+  (strict-failure pre-decision, retry-exhausted with a `fallback_retry_target`)
+  keep the warning-only behavior so the routing outcome is never changed. No
+  behavior change when devices and the artifact repo are healthy.
 
 ## [0.40.2] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Sandbox device-node hygiene preflight (#423).** Before any git or subprocess
+  handler runs — on both fresh runs and resume — `applyGitPreflight` now verifies
+  standard device nodes are usable (at minimum `/dev/null` is a readable+writable
+  character device). A suspended/restored sandbox can silently corrupt `/dev/null`
+  (e.g. it becomes unreadable or a regular file masquerading as the device),
+  which breaks git and reviewer CLIs deep mid-run; that previously surfaced only
+  as an opaque `context canceled` failure during the terminal commit. The probe
+  now fails fast with a specific, actionable diagnostic (including a `mknod`
+  remediation) instead. The probe is portability-guarded (POSIX real check,
+  Windows no-op) and injectable, so it runs ahead of and independent of the git
+  policy and never touches the host device in tests. No new behavior when
+  `/dev/null` is healthy.
+
+### Changed
+
+- **Artifact-repo health check + reattach on the never-lose-work commit path
+  (#423).** `commitWIPBeforeRouting` now probes artifact-repo availability via a
+  new `gitArtifactRepo.ensureHealthy()` before committing work-in-progress; if
+  the repo has gone unreachable post-suspend it attempts a single reattach
+  (clear the latched failure + idempotent re-`Init`, capturing the current tree).
+  On the **terminal** never-lose-work paths (handler-error halt and failing exit
+  node) an unrecoverable repo is now surfaced as a HARD signal — a new
+  `EventStageFailed` diagnostic plus an `EngineResult.WorkPreserveFailed` flag —
+  rather than degrading silently to a best-effort `EventWarning`. The original
+  node failure is never masked (`Status` stays `OutcomeFail`). **Mid-routing**
+  call sites (strict-failure pre-decision, retry-exhausted with a
+  `fallback_retry_target`) keep the warning-only behavior so the routing outcome
+  is never changed. No behavior change when devices and the artifact repo are
+  healthy.
+
 ## [0.40.2] - 2026-06-24
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   new `gitArtifactRepo.ensureHealthy()` before committing work-in-progress; if
   the repo has gone unreachable post-suspend it attempts a single reattach
   (clear the latched failure + idempotent re-`Init`, capturing the current tree).
+  The health probe compares `git rev-parse --show-toplevel` against the artifact
+  dir, so a nested run dir whose own `.git` was lost is treated as unhealthy
+  (and reattached) instead of silently resolving against — and committing into —
+  the user's enclosing repository.
   On the **terminal** never-lose-work paths (handler-error halt and failing exit
   node) an unrecoverable repo is now surfaced as a HARD signal — a new
   `EventStageFailed` diagnostic plus an `EngineResult.WorkPreserveFailed` flag —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `EventWarning` (the mid-routing path emits the single discarded warning at its
   own site); and the device-probe remediation hint no longer presents the
   Linux-specific `mknod ... c 1 3` device numbers as if they were portable.
+  Follow-up review fixes (#428): the strict-failure mid-routing fallback site
+  (`strictFailureFallback`) now actually emits that discarded-preserve
+  `EventWarning` — it was previously dropped silently, leaving only the
+  retry-exhausted fallback branch warning; and `checkRootedHere` now includes
+  git's combined output in the `rev-parse --show-toplevel` error so a
+  repo-unavailable diagnostic stays actionable (e.g. surfaces `safe.directory`
+  or corrupt-object causes) instead of reporting a bare exit status.
 
 - **Sleep-aware budgets (#422, part A).** Opt-in `BudgetLimits.SleepAware` (CLI
   `--sleep-aware-budget`) excludes OS-suspend spans — e.g. a suspended laptop —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   masked (`Status` stays `OutcomeFail`). **Mid-routing** call sites
   (strict-failure pre-decision, retry-exhausted with a `fallback_retry_target`)
   keep the warning-only behavior so the routing outcome is never changed. No
-  behavior change when devices and the artifact repo are healthy.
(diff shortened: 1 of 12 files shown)


clintecker:
Addressed the 06-29 review round on `gitArtifactRepo.ensureHealthy`:

**Codex (P1) — reject parent-repo discovery:** `git rev-parse --git-dir` walks *up* the directory tree, so a nested artifact run dir whose own `.git` was lost resolved against the user's enclosing repository — the probe declared the repo healthy and skipped reattach, and subsequent WIP/node commits could have mutated the wrong repo. Replaced the probe with `checkRootedHere()`, which compares `git rev-parse --show-toplevel` (symlink-resolved via `resolveSymlinksOrFallback`, case-folded on Windows via `samePathForLatch`) against the artifact dir. Parent discovery is now treated as unhealthy and triggers reattach to a fresh repo rooted at the nested dir. New test `TestEnsureHealthy_RejectsParentRepoDiscovery` builds a nested repo under a parent git repo, deletes the nested `.git`, asserts parent discovery as a precondition, then asserts reattach re-roots `--show-toplevel` back at the nested dir.

**Copilot — `%v` on inner reattach error:** Changed to `%w` so callers can `errors.Is`/`errors.As` the underlying `Init` failure cause.

**Copilot — `: <nil>` in the post-reattach probe:** `checkRootedHere()` returns a descriptive, never-nil error (distinct messages for git error, empty toplevel, and parent discovery); the caller wraps it with `%w` alongside `ErrArtifactRepoUnavailable`. No more misleading `: <nil>` suffix.

**Copilot — unix-only test:** Added `//go:build unix` to `pipeline/git_artifact_health_test.go` (it uses `os.Geteuid()` + `chmod 0500`), matching the repo's other unix-only tests.

`go build`, `go test -short`, race detector, coverage, complexity, and dippin lint all green via the pre-commit gates.

clintecker:
## Addressed latest Copilot review feedback (commit cd984f1)

**Finding: terminal work-preserve halt was emitted as an extra `stage_failed` event with the same NodeID as the node failure, inflating `tracker diagnose`'s per-node `RetryCount` / `IdenticalRetries`.**

The terminal never-lose-work HARD signal (#423) was previously emitted via `EventStageFailed`. `tracker diagnose` counts `stage_failed` events per NodeID toward that node's retry/failure stats — but the preserve-failed diagnostic is *not* another execution attempt for the node, so it was double-counting.

Fix:
- New dedicated event type **`EventWorkPreserveFailed`** (`pipeline/events.go`). `escalateWorkPreserve` now emits it instead of `EventStageFailed` (`pipeline/engine_run.go`).
- `tracker diagnose`'s `stage_failed` case no longer matches it → retry/failure counts are correct.
- TUI adapter (`tui/adapter.go`) maps `EventWorkPreserveFailed` into the **same** `MsgNodeFailed` failure line (collapsed into the existing `EventStageFailed` case), so the operator still sees the hard signal — TUI behavior is byte-identical.
- Tests in `git_artifact_health_test.go` updated to assert the new event type on the terminal path (and its absence on the mid-routing warning path).
- CHANGELOG `#423` entry updated to document the distinct event type and the rationale.

Incidental: staging `tui/adapter.go` brought pre-existing `AdaptAgentEvent` (cyclomatic 10) under the whole-file complexity gate; split its tool/status cases into `adaptAgentToolEvent` to satisfy the gate (no behavior change).

Gates: `go build`, `go test -short`, race detector, `pipeline` coverage 85.2%, complexity ≤ 8, dippin lint — all green.

clintecker:
## Round-2 review feedback addressed (ffdccec) + main merged (404f5ec)

**New review findings:**

1. **Stale comment in `pipeline/engine.go` (terminal never-lose-work path).** The doc block claimed the degradation was surfaced via `EventWarning`; it now correctly states `escalateWorkPreserve` emits the dedicated `EventWorkPreserveFailed` and sets `EngineResult.WorkPreserveFailed`. Comment-only; behavior already matched the corrected text.

2. **Nil-deref risk in `pipeline/git_artifact_health_test.go` cleanups.** All four `t.Cleanup` closures `os.Chmod`'d `filepath.Join(artifactBase, result.RunID)` unconditionally — if the run errored before `result` was assigned, the cleanup would panic on a nil deref and mask the real failure. Each closure now guards `if result != nil { … }`.

**Merge:** rebased onto `origin/main` (which now carries #426 sleep-aware budgets). Two conflicts resolved:
- `CHANGELOG.md` `[Unreleased]` — kept both #428 (sandbox device hygiene + artifact-repo health) and #426 (sleep-aware budgets) entries.
- `pipeline/engine.go` — both branches independently extracted a `successResult` helper; the textual auto-merge produced a self-calling stub plus a duplicate method. Consolidated into the single self-contained `successResult` (finalizes trace, emits `EventPipelineCompleted`, builds override-aware terminal result).

All pre-commit gates green (build, vet, test -short, race, coverage, complexity, dippin lint).

clintecker:
## Round-3 review feedback addressed (e57698d)

1. **Double diagnostic on terminal halt** (engine_run.go:80). On a terminal never-lose-work halt with an unrecoverable artifact repo, the run emitted *both* an `EventWarning` (from `commitWIPBeforeRouting`'s `ensureHealthy` failure) and the hard `EventWorkPreserveFailed` (from `escalateWorkPreserve`) — same node, same cause. **Fix:** `commitWIPBeforeRouting` no longer emits a diagnostic when `ensureHealthy` fails; it just returns the error and lets the caller own severity. Terminal paths hard-escalate **once** (`EventWorkPreserveFailed`); the lone mid-routing discard site (`handleRetryExhausted` fallback branch) now emits a single `EventWarning` itself — so no path is silent and none double-logs.

2. **Non-portable device remediation hint** (device_probe.go:30). The hint hard-coded `mknod ... c 1 3` (Linux device numbers) but the message is shown on all non-Windows platforms. Reworded to `Recreate /dev/null (on Linux: \`mknod -m 666 /dev/null c 1 3\`; device numbers differ on other platforms)`. The `device_probe_test` assertion on `"mknod"` still holds.

**Already addressed in prior commits (re-flagged on stale revisions):**
- t.Cleanup nil-deref (git_artifact_health_test.go:201/284/386/429) — all four guarded with `if result != nil` (ffdccec).
- `//go:build unix` build tag on the health test — present.
- engine.go comment saying `EventStageFailed` — corrected to `EventWorkPreserveFailed` (ffdccec).
- `escalateWorkPreserve` emitting `stage_failed` and inflating RetryCount — that is exactly why it uses the dedicated `EventWorkPreserveFailed` event type.
- git_artifacts.go `%w` wrapping + non-nil reattach-probe diagnostics — `ensureHealthy`/`checkRootedHere` already use `%w` and descriptive errors.

All pre-commit gates green (build, vet, test -short, race, coverage, complexity, dippin lint).

clintecker:
## Latest Copilot round addressed (commit 9a28644)

**1 & 2. `breakRepoHandler` / `breakRepoOnStatusHandler` ignored `ok` from `GetInternal` and discarded file-op errors** (`git_artifact_health_test.go:156`/`:316`). If the internal artifact dir were ever missing/empty, `filepath.Join("", ".git")` resolves to a relative `.git` and `os.RemoveAll` could delete the `.git` of the repo checkout the test runs in. Extracted a shared `breakArtifactRepo` helper that:
- refuses to act unless the artifact dir is a non-empty **absolute** path (`!ok || dir == "" || !filepath.IsAbs(dir)` → `t.Errorf` + return, no destructive op);
- surfaces every `WriteFile`/`RemoveAll`/`Chmod` error via `t.Errorf` (safe from a handler goroutine — it calls `Fail`, not `FailNow`) instead of `_ =` discarding it, so a broken precondition fails the test loudly rather than silently invalidating the signal.

Both handlers now call it, removing the duplication.

**3. `device_probe_test.go:56` cleanup clobbered global state.** The `TestApplyGitPreflight_DeviceCheckedFirst` cleanup reset `activeGitConfig.policy` to `""` rather than its prior value, leaking state into other `cmd/tracker` tests. Now captures `origPolicy` and restores it.

All pre-commit gates green (build, vet, 19-pkg tests, race, coverage, complexity, dippin lint).

clintecker:
Addressed the two genuinely-new Copilot findings from this round (commit 4812cdf):

**1. `pipeline/engine.go` — strict-failure mid-routing fallback silently discarded the preserve error.**
`strictFailureFallback` routed an unhandled strict failure to its `fallback_target` but dropped the `commitWIPBeforeRouting` error without surfacing it. The surrounding comment and the CHANGELOG already claimed the mid-routing path emits "a single discarded warning at its own site," but in practice only `handleRetryExhausted`'s fallback branch emitted one — this branch was silent, a never-lose-work degradation going unreported. Fixed by threading `preserveErr` into `strictFailureFallback` and emitting one `EventWarning` when it is non-nil, mirroring the retry-exhausted branch. Routing is unchanged: the terminal branch in `checkStrictFailure` still hard-escalates via `EventWorkPreserveFailed`; only the mid-routing path warns (so escalation can't override the routing decision).

**2. `pipeline/git_artifacts.go` — `checkRootedHere` dropped git's output on `rev-parse` failure.**
The `rev-parse --show-toplevel` error reported only the bare exit status, discarding git's combined stdout/stderr. Fixed to append `\n%s` with `strings.TrimSpace(out)`, matching every other error site in the file, so a repo-unavailable diagnostic (`ErrArtifactRepoUnavailable`) stays actionable — surfacing `safe.directory`, corrupt-object, or bad-ref causes instead of an opaque code.

Added `TestStrictFailureFallback_MidRoutingStaysWarning`: a broken artifact repo on a strict-failure node with a `fallback_target` asserts a single `EventWarning` surfaces, routing still reaches the fallback, and `WorkPreserveFailed` stays false.

Full pre-commit gates green (build, vet, 19-pkg tests, race, pipeline coverage 85%+, complexity, dippin lint).

clintecker:
Addressed the new Copilot finding (commit 5d3a36f):

**`cmd/tracker/device_probe_posix.go` — ABOUTME said "POSIX" while the build tag is `!windows`.** `!windows` is broader than POSIX (it also matches `js/wasm`, `wasip1`). Reworded the ABOUTME to "non-Windows real implementation" to match the actual tag, and added a note that the `_posix` filename is descriptive only (Go assigns no build meaning to a `_posix` suffix — only recognized GOOS/GOARCH suffixes are magic). I kept the `!windows` tag rather than tightening to `unix` + adding a stub, since tracker only ships darwin/linux (GoReleaser targets) and there is no non-Windows target where `/dev/null` probing is undesirable in practice — so no build-coverage gap is introduced.

Full pre-commit gates green.

PR #512, opened 2026-07-28
feat(pipeline): bounded/async event-handler seam with explicit overflow policy

clintecker:
## Problem

`tracker_events.go` documents the hazard in its own comments: event handlers are
invoked **synchronously on the engine goroutine**. A subscriber doing network I/O
(a control plane POSTing events — exactly the shape the downstream
`tracker-runner` product has) blocks the engine, and three handler sources
sharing one writer serialize across each other. The documented mitigation was
"wrap it yourself in a `bufio.Writer` or a channel-backed forwarder", so every
embedder re-implements the same bounded queue, with overflow policy and drop
accounting left to chance.

## Design

New `pipeline/events_buffered.go` ships it once: `BufferedPipelineHandler` and
`BufferedAgentHandler` hand events to a background goroutine over a bounded
queue, sharing a generic `eventQueue[T]`.

Public API:

```go
type OverflowPolicy string
const (
    OverflowBlock      OverflowPolicy = "block"
    OverflowDropOldest OverflowPolicy = "drop_oldest"
    OverflowDropNewest OverflowPolicy = "drop_newest"
)

func NewBufferedPipelineHandler(inner PipelineEventHandler, capacity int, policy OverflowPolicy) (*BufferedPipelineHandler, error)
func (h *BufferedPipelineHandler) HandlePipelineEvent(evt PipelineEvent)
func (h *BufferedPipelineHandler) Dropped() uint64
func (h *BufferedPipelineHandler) Close() error

func NewBufferedAgentHandler(inner agent.EventHandler, capacity int, policy OverflowPolicy) (*BufferedAgentHandler, error)
func (h *BufferedAgentHandler) HandleEvent(evt agent.Event)
func (h *BufferedAgentHandler) Dropped() uint64
func (h *BufferedAgentHandler) Close() error
```

- **Overflow policy is explicit.** `OverflowBlock` (backpressure, drops nothing),
  `OverflowDropOldest` (freshest view; progress UIs), `OverflowDropNewest`
  (earliest prefix). The type is string-backed so the zero value is unusable —
  an unset policy, a nil inner handler, or `capacity < 1` is a constructor
  error. No implicit default, because every default either silently loses data
  or silently applies backpressure and the caller must own that choice.
  `Dropped()` counts overflow drops plus non-terminal post-`Close` events, so
  every submitted event is either delivered or counted.
- **Terminal-event invariant:** an event with a non-empty `TerminalStatus` is
  **never dropped under any policy** — it is the run-finished signal
  `docs/architecture/transport-boundary.md` promises subscribers, and dropping
  it strands a consumer forever. At a full queue a terminal event evicts the
  oldest *non-terminal* event (bounded, non-blocking); only a queue holding
  nothing but undelivered terminal events applies backpressure. `OverflowBlock`
  never evicts anything. A terminal event submitted after `Close` is delivered
  synchronously, after the flush, rather than dropped.
- **`Close()`** flushes pending events, stops the goroutine, is idempotent and
  safe to call concurrently with sends. Every channel operation happens under one
  mutex guarded by a `closed` flag, so send-on-closed-channel is impossible; a
  blocking send resolves because the forwarding goroutine keeps draining until
  the channel is closed.
- **Panic safety:** a panicking downstream handler is recovered and logged once
  to stderr (following the `recoverPanic` precedent in `tracker_events.go`); the
  forwarding goroutine survives and the engine is untouched.

Docs: new "Don't block the engine: buffered handlers" subsection in
`docs/architecture/transport-boundary.md` §3, with a usage snippet. `CHANGELOG.md`
updated under `## [Unreleased]` → `### Added`. No existing files touched beyond
those two.

## Verification (actually run)

| Command | Result |
|---|---|
| `go build ./...` | pass |
| `go test ./pipeline/... -short -count=1` | `ok pipeline 117.9s`, `ok pipeline/handlers 2.5s` |
| `go test ./pipeline/ -run Buffered -race -count=5` | `ok pipeline 1.6s` |
| `make complexity` | `complexity gate OK (212 grandfathered, 0 new)` |
| `gofmt -l pipeline/`, `go vet ./pipeline/` | clean |
| pre-commit hook (format/vet/build/tests/complexity/dippin lint) | all gates passed |

Tests pin: a slow (blocked) downstream handler does not stall the producer
(pipeline + agent); drops are counted under each policy and
delivered+dropped == submitted; `OverflowBlock` drops nothing even when a
terminal event hits a full queue; a terminal event survives overflow under all
three policies; two terminal events survive a saturated queue; a post-`Close`
terminal event is delivered while a post-`Close` non-terminal is counted;
`Close` flushes and is idempotent (and non-deadlocking under 8 concurrent
producers × 3 concurrent `Close`s); a panicking handler is contained and all
subsequent events still deliver; constructor validation.

## Self-review found

1. **`OverflowBlock` could drop.** The terminal-event branch ran before the
   policy check, so a terminal event arriving at a full `OverflowBlock` queue
   evicted (and counted as dropped) a non-terminal event — contradicting
   "backpressure drops nothing". Reordered the switch and added
   `TestBufferedPipelineHandlerBlockPolicyDropsNothingWhenFull`, verified to fail
   against the old ordering.
2. **Post-`Close` terminal delivery could race the flush.** Inline delivery now
   waits for the forwarding goroutine to exit first, so the terminal event lands
   after everything `Close` flushed and never concurrently with it. (An earlier
   draft used a delivery mutex held by the forwarder, which deadlocked a producer
   against a wedged downstream handler; dropped in favour of the wait.)
3. Documented the residual re-entrancy hazard (a wrapped handler must not call
   back into the wrapper) and that `Close` waits for the flush, so an unbounded
   subscriber keeps `Close` waiting.

## Deliberately left out

- No wiring into `Config` / `tracker.go` / the CLI — the item is the library
  seam, and those files are being edited concurrently.
- No timeout/deadline option on the downstream handler (a wedged subscriber
  still wedges `Close`) and no metrics hook beyond `Dropped()`.
- `agent.Event` has no terminal status, so `BufferedAgentHandler` subjects every
  event to the policy; the invariant is pipeline-only by construction.

https://claude.ai/code/session_01RuhYJxVgQPsfuHZpAYbeed

---
<a href="https://app.blacksmith.sh/2389-research/codesmith/tracker/pr/512"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787860512&installation_model_id=21126&pr_number=512&repository=2389-research%2Ftracker&return_to=https%3A%2F%2Fgithub.com%2F2389-research%2Ftracker%2Fpull%2F512&signature=7f42b2cfca0e1dbeb4cc9ddf327a8e6ae24addf8e6a82a1c743e00793be42db7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added bounded, asynchronous event handlers for pipeline and agent events.
  - Added configurable overflow behavior: block, drop oldest, or drop newest.
  - Added dropped-event counters and graceful, idempotent shutdown with queue flushing.
  - Ensured terminal events are always delivered, even when buffers are full or closing.
  - Contained downstream handler panics so event processing continues.

- **Documentation**
  - Added guidance for using buffered handlers with slow or network-based processing.


clintecker:
Both findings were real — evaluated, reproduced with failing tests, and fixed in e93e297 (branch also rebased onto current `main` after #511 merged). No pushback; the reviewer read the terminal-protection path correctly.

## P1 — backpressure applied even when a non-terminal event was evictable

Confirmed. `evictOldest` examined only the head: a terminal event there was rotated to the tail and the function returned `false`, which the terminal branch of `handleFull` ignored before blocking on the send. So a queue of `[sub-budget terminal, stage, stage]` stalled the producer behind a slow sink under **both** drop policies with two evictable non-terminals sitting right there — and that arrangement needs nothing exotic: a scoped subgraph `budget_exceeded` followed by more stage events in the same run produces it.

Fix: `evictOldest` now rotates protected terminals to the tail and keeps searching (bounded by `cap(ch)`), reporting backpressure only when *every* queued event is terminal — the one case where blocking is the sole alternative to dropping a run-finished signal. Rotation is the single place the wrapper reorders a stream (a protected terminal can land after later non-terminals); terminal events keep their order relative to each other and are never dropped. That caveat is now stated in the code, the `transport-boundary.md` subsection, and the CHANGELOG.

Regression test `TestBufferedPipelineHandlerEvictsPastQueuedTerminal` builds exactly the head-is-terminal-with-non-terminal-behind queue (a `parkingSink` parks the forwarding goroutine first so the queue contents are deterministic, not timing-dependent). Before the fix:

```
--- FAIL: TestBufferedPipelineHandlerEvictsPastQueuedTerminal (4.00s)
    --- FAIL: .../drop_oldest (2.00s)
        producer stalled behind a wedged sink even though an evictable non-terminal event was queued
    --- FAIL: .../drop_newest (2.00s)
        producer stalled behind a wedged sink even though an evictable non-terminal event was queued
```

`TestBufferedPipelineHandlerAllTerminalQueueAppliesBackpressure` guards the complementary invariant so the new search can't regress into evicting a terminal: an all-terminal queue drops nothing and both terminals still land.

## P2 — concurrent post-`Close` dispatches raced, and `Close` could return mid-write

Confirmed on both halves. Every post-`Close` terminal observed the completed `WaitGroup` and called `dispatch` directly, so several could enter the sink at once (contradicting my own documented invariant), and `Close` returned as soon as the forwarding goroutine exited while such a dispatch was still writing.

Fix: post-`Close` delivery takes a dedicated `postCloseMu` (as proposed), **plus** lifecycle tracking — `enqueue` registers the dispatch in an `inflight` counter under the existing `mu`, and `close` waits on a `sync.Cond` until it drains. I used a counter + `Cond` rather than a second `WaitGroup` deliberately: a `WaitGroup.Add` from a producer racing `Close`'s `Wait` is the documented Add-during-Wait misuse. An event submitted *after* `Close` has already returned remains the caller's ordering problem — the doc comment now says that explicitly instead of implying a guarantee the code cannot make.

Before the fix:

```
--- FAIL: TestBufferedPipelineHandlerPostCloseTerminalDispatchIsSerialized (0.02s)
        wrapped handler entered concurrently 7 times; post-Close terminal dispatch must be serialized
--- FAIL: TestBufferedPipelineHandlerCloseWaitsForPostCloseDispatch (0.00s)
        Close returned while a post-Close terminal dispatch was still writing to the sink
```

## Docs

The stale "single goroutine ... with one exception" comment is replaced by what the code now actually guarantees: delivery is serialized, the wrapper never invokes the sink from two goroutines at once, and `Close` leaves the sink quiet. The `transport-boundary.md` §3 subsection and CHANGELOG entry were updated to match, including the rotation/reordering caveat.

## Verification (re-run after the fixes)

| Command | Result |
|---|---|
| `go build ./...` | pass |
| `go test ./pipeline/... -short -count=1` | `ok pipeline 57.8s`, `ok pipeline/handlers 1.0s` |
| `go test ./pipeline/ -run Buffered -race -count=5` | `ok pipeline 4.1s` |
| `make complexity` | `complexity gate OK (212 grandfathered, 0 new)` |
| `gofmt -l pipeline/`, `go vet ./pipeline/` | clean |
| pre-commit hook (format/vet/build/22-package tests/complexity/dippin lint) | all gates passed |

New tests live in `pipeline/events_buffered_review_test.go` so `events_buffered.go` stays at 323 lines and `events_buffered_test.go` at 423 — both under the 500-line ratchet.

Issue #504, opened 2026-07-26
.dip/engine: multi-operator, multi-condition edge predicates (contains/compare/regex, AND/OR, else) beyond ctx=value equality


clintecker:
## Summary
`.dip` edges currently route only on single-key **string equality** — `A -> B when ctx.KEY = VALUE`. There is no operator set, no multi-condition combination, and no numeric comparison, so a pipeline that already `writes:` context flags cannot branch on them and authors fake multi-way routing with chained success/fail retries.

## Evidence
- tracker-runner pipelines: 12/12 edges route on `ctx.outcome ∈ {success, fail}` (`pipelines/*.dip:132-135`); `writes:` values never appear in a routing predicate.
- Harvested from **langgenius/dify** (`if_else`/switch: condition *lists* with `contains / not_contains / starts_with / is_empty / is_null / in / >,<,>=,<= / regex`, combined with `and`/`or`, each case routing to its own branch, plus an explicit `else`) and corroborated by **bytedance/deer-flow** (`add_conditional_edges(..., [N allowlisted targets])` — routing on arbitrary state, not just pass/fail). See `docs/harvest/dify-harvest.md` (A) and `docs/harvest/deer-flow-harvest.md` (④) in tracker-runner.

## Proposal
Extend the `.dip` edge grammar + engine edge-selection to support:
- Comparison operators (`contains`, `starts_with`, `is_empty`, `in`, `>`, `<`, `>=`, `<=`, `regex`) beyond `=`.
- Multiple conditions combined with `and`/`or`.
- An explicit `else`/default branch, with N allowlisted targets from one node.

## Acceptance criteria
- An edge can route on a non-`outcome` context key using a comparison operator.
- A classify node can fan to ≥3 distinct branches by a single review/verdict value.
- Ambiguous/unmatched routing falls to an explicit default (fail-closed, not silent drop).
- Depends on / pairs with the typed-variable-validation issue (do that first so numeric compares are sound).

_Filed from a 2026-07-26 external source-audit sweep (tracker-runner `docs/harvest/`)._

clintecker:
**Scope update (verified against main @ 983cdd8):** most of this is already implemented — the premise "beyond ctx=value equality" is stale.

`pipeline/condition.go` (`EvaluateCondition`) already supports, and edge selection already routes through it (`pipeline/engine_edges.go:95`):
- **contains** / **startswith** / **endswith** / **in** (+ their `not …` negations)
[...]

PR #510, opened 2026-07-28
fix(events): turn_metrics attribution + gate lifecycle events (#508, #509)

clintecker:
Closes #508. Closes #509.

Two event-stream gaps reported by **tracker-runner**, an embedded consumer that builds cost rollups and gate history from the stream alone.

## #508 — `turn_metrics` carried no attribution

The event was emitted with an empty `Model`/`Provider` and zero `Usage`. A consumer reading per-turn usage from `turn_metrics` (a reasonable reading of the name) silently got zeros and concluded the run was free — the real numbers were only on `llm_finish`.

`turn_metrics` now carries the same top-level attribution as `llm_finish`: taken from the LLM response, falling back to the session's configured model/provider for adapters that leave the response fields unset. The per-turn `Metrics` payload is unchanged.

## #509 — no gate lifecycle events existed

There were no gate events at all on tracker's stream, so a pure event-stream consumer could not tell which gate node opened, what was asked, or which answer belonged to which question.

The human gate handler now emits `gate_opened` / `gate_resolved` around every interviewer call, carrying a `GateDetail` on `PipelineEvent.Gate` with `NodeID` set to the gate node:

| | fields |
|---|---|
| `gate_opened` | `gate_id`, `mode`, `label`, `prompt` (capped at `GateMaxPromptBytes`), `choices` |
| `gate_resolved` | same `gate_id`, plus `response`, `outcome`, `actor`, `timed_out`, error reason |

Exactly one resolution follows each open — including on failure, timeout, and interviewer error — so a gate is never left stuck open. Fields land on `activity.jsonl` as `gate_*`; the NDJSON `StreamEvent` wire format gains `gate_id` (additive, empty on every non-gate event). Handlers constructed directly opt in via `handlers.WithHumanPipelineEmitter`; a nil emitter is previous behavior exactly.

## Mechanically required, no logic change

The pre-commit hook enforces a hard cyclo/cognitive ≤8 on *staged* files (unlike the baseline-aware CI ratchet), so touching these files forced burning down their grandfathered debt:

- decomposed `session.go` `Run`/`runTurnLoop`/`turnHasEdits`/`runVerifyLoop` and `events_jsonl.go` `buildLogEntry`/`WriteAgentEvent`/`writeSnapshot`
- split three files that would otherwise exceed their file-size baseline: `agent/session_events.go`, `pipeline/events_jsonl_snapshot.go`, `pipeline/handlers/{human_gate_events,interviewer_console}.go` (pure moves)
[...]

Issue #509, opened 2026-07-28
Gate events under-specified: gate_opened lacks node_id/label, gate_resolved lacks gate_id


clintecker:
**Observed in v0.46.0** running a pipeline with 5 human gates (freeform + choice modes). From the event stream:

```json
{"event_type":"gate_opened",   "payload":{"gate_id":"gate_20b979c8b8...","mode":"freeform"}}
{"event_type":"gate_resolved", "payload":{"resolved_by":"U08FWCAA13J","response":"Birds Are Real"}}
```

Two gaps for event-sourced consumers:
1. **`gate_opened` has no `node_id` and no prompt/label** — you can't tell which gate node opened, or what question was asked, from the stream.
2. **`gate_resolved` has no `gate_id`** — a resolution can't be correlated back to the gate that opened. With 5 gates in one run, reconstructing 'which question got which answer' from the stream alone is not possible.

**Impact: low for us** — our control plane owns the gate lifecycle (it supplies the prompt when creating the gate and stores node_id/prompt/choices itself), so we're not blocked. It matters more for a pure event-stream consumer, a replay/audit tool, or anyone rendering gate history from events.

**Ask:** add `node_id` (and ideally the gate label) to `gate_opened`, and `gate_id` to `gate_resolved`.

PR #513, opened 2026-07-28
feat(validate): submit-time variable-availability validation (#505)

### LINKED ISSUE #505: .dip/engine: typed node variables + submit-time variable-availability validation (reject references to unwritten ctx keys)
## Summary
`.dip` nodes pass **free-text `writes:` values** with no schema. Nothing validates at submit time that a `when ctx.x` reference actually resolves to a variable some upstream node writes — so a typo or a mis-wired graph is an inert-wiring bug discovered only at runtime (or never). Free-text values also block sound numeric comparison in edge predicates.

## Evidence
- tracker-runner `internal/trackerbridge/validate.go` (`ValidatePipelineSource`) validates structure but not variable reachability; `inspectGraph` already walks the graph.
- Harvested from **langgenius/dify**: `WorkflowGraphTopology.is_upstream / upstream_node_ids` validates at publish that every referenced variable is produced by an upstream node. See `docs/harvest/dify-harvest.md` (B).

## Proposal
- Let nodes declare typed I/O variables (at least: string / number / bool / json).
- At pipeline submit/validate, assert every `ctx.<key>` referenced in an edge predicate or a prompt/`reads:` is produced by a node upstream of the reference; fail submission otherwise.
- Surface the failing (node, key) pair.

## Acceptance criteria
- A pipeline referencing an unwritten/misspelled context key is **rejected at submit**, with the offending node+key named (observable-effect, not a runtime surprise).
- A typed numeric variable can be compared with `>`/`<` in an edge predicate (enables the edge-operators issue).
- Existing valid pipelines still validate.

_Filed from a 2026-07-26 external source-audit sweep (tracker-runner `docs/harvest/`). Do this before the multi-operator-edges issue._

clintecker:
Closes #505.

## Problem

A `.dip` pipeline can reference a context key that no upstream node ever writes — `when ctx.typo_key = success`, a `${ctx.mispelled}` interpolation, a `reads:` of a key nothing produces. Nothing caught it at validate/submit time, so a typo or a mis-wired graph was found either at runtime (a condition silently evaluating empty, routing the wrong way) or never. The downstream `tracker-runner` control plane validates pipeline *structure* before accepting a submission but could not prove variable reachability, so it accepted pipelines it could have proven broken.

## What this adds

`pipeline.ValidateVariableAvailability(g)` (new `pipeline/validate_vars.go` + `pipeline/validate_vars_refs.go`) walks the graph and reports every `ctx.<key>` reference — edge `when` predicates, `${ctx.*}` interpolations in node attributes, and declared `reads:` entries — that no node in the pipeline can produce, naming the offending `(node, key)` pair.

Wired in at two seams:
- `ValidateSemantic` → `tracker validate` / `simulate` / `doctor`
- `tracker.ValidateSource` → the library submit seam a control plane calls (it uses `ValidateAll`, which takes no handler registry, so the registry-independent check is invoked directly)

The **run** path (`pipeline.Validate`, used by `tracker run` and `tracker.Run`) is deliberately untouched — this is a submit-time gate, and embedders that inject keys via `Config.Context` must not have runs broken by it.

## Reachability rule

**Plain directed-graph reachability** (transitive closure over edges), **not** topological order. A producer P can reach a consumer C when some path `P -> ... -> C` exists, and every node is treated as reaching itself.

Cycles are therefore first-class: a fix-loop or a `manager_loop` that writes a key on one iteration legitimately makes it available to the nodes it loops back to, and a node that writes a key and also reads it is legal on a second visit. A topological / strictly-upstream rule would flag every loop in `examples/build_product.dip` — this rule flags none of them.

## Fail-open default (and why)

Rejecting a valid pipeline is far worse than missing an invalid one, so **only the unambiguous case is an error**: the key has *no* producer anywhere in the graph **and** the reference is one the author clearly meant as a context variable (an explicit `ctx.`/`context.`-prefixed condition operand, or a declared `reads:` entry).

Everything softer is a **warning**:
- a producer exists but cannot reach the reference — engine restarts, `restart_target`, checkpoint resume and `suggested_next_nodes` all move control along paths the static edge set does not express;
- a bare (unprefixed) condition operand — it may be a literal, not a key;
- a `${ctx.*}` interpolation — it expands to the empty string at runtime rather than mis-routing, and it is where externally-injected values realistically appear;
- an unknown node id inside `node.<id>.<key>`.

**Never flagged at all:**
- the engine/handler-provided key set (`outcome`, `preferred_label`, `last_response`, `human_response`, `tool_stdout`/`tool_stderr`/`tool_marker`/`tool_marker_error`/`tool_route`, `suggested_next_nodes`, `turn_limit_msg`, `turn_breach_class`, `node_cost_exceeded`, `node_no_progress`, `episode_summary`/`episode_summaries`, `interview_questions`/`interview_answers`, `last_cost`, `last_turns`, `writes_error`, `writes_warning`, `completed_summary`, `graph.goal`);
- every dynamic namespace: `graph.`, `params.`, `response.`, `node.`, `steer.`, `stack.`, `summary.`, `parallel.`, `fan_in.`, `internal.`, `_*`;
- any reference reachable from an **opaque producer** — `subgraph` and `stack.manager_loop` nodes merge back keys the enclosing graph cannot enumerate, so their whole downstream reach is skipped.

Producers indexed positionally are only the *author-declared* ones: `writes:` keys and a human interview node's `answers_key`. Keeping handler built-ins out of positional reasoning removes that entire false-positive class.

**Known limitation (documented in code, CHANGELOG and `docs/architecture/context-flow.md`):** keys injected from outside the graph — the library's `Config.Context`, or a parent's context snapshot handed to a subgraph child — are invisible to the walk. The fix is to name the key in the producing node's `writes:` so the contract lives in the graph that consumes it.

## Verification (all actually run)

- `go build ./...` — pass
- `go test ./... -short -count=1` — pass (22 packages, also via the pre-commit gate)
- `make complexity` — `complexity gate OK (212 grandfathered, 0 new)`. Two rounds of decomposition were needed: `conditionVarOperands` was split (`branchVarOperands`), and the implementation was split across two files after the single file crossed the 500-line ratchet.
- **Every embedded example still validates.** `TestValidateVariableAvailability_ExamplesClean` walks all 38 `examples/**/*.dip` and asserts **zero errors and zero warnings** from this check. Also ran `tracker validate` over every example: no new errors and no new warnings (the warnings shown are pre-existing DIP1xx / TRK102 lint). The one pre-existing failure, `examples/manager_loop_demo.dip: node "Supervise" has unregistered handler "stack.manager_loop"`, is unrelated (`buildValidationRegistry` registers `manager_loop`, not `stack.manager_loop`) and unchanged by this PR.
- `dippin doctor` (dippin is on PATH; **not** installed by this work): `ask_and_execute.dip` **A** 100/100, `build_product.dip` **A** 90/100, `build_product_with_superspec.dip` **A** 100/100 — identical before and after.
- Required tests present: a misspelled key is rejected with an error naming both node and key; every engine-provided key and a loop-written key are asserted un-flagged; plus self-writes, unreachable producers, opaque nodes, `answers_key` producers, scoped-ref node ids, and one-finding-per-(node,key) deduplication.

## Self-review pass

Re-read the whole diff hunting false-positive sources. What it caught and fixed:
- **Duplicate findings.** The same missing key declared in `reads:` *and* interpolated produced one error plus N warnings. Added `(node, key)` deduplication that keeps the fatal finding (`upsertVarFinding`), with a test.
- **`answers_key` producers.** `examples/deep_review.dip` reads `${ctx.scope_answers}` / `${ctx.findings_answers}` / `${ctx.priority_answers}`, which a human interview node writes via `answers_key:`. Without indexing that attr, three false positives. Added.
- **`stack.*` and `steer.*`.** `examples/manager_loop_demo.dip` reads `${ctx.stack.child.status}` / `.cycles` and `${ctx.steer.hint}`. Added both to the ambient prefixes.
- **Parallel branch overrides.** Checked whether `branch.N.*` per-branch attrs can override `writes:` (they cannot — only model/provider/fidelity/tool_access/writable_paths), so no gap there. Branch writes reach post-fan-in consumers through the synthesized `parallel_targets` / `FanInConfig.Sources` edges, which plain reachability already covers.
- **Externally-injected keys.** Identified `Config.Context` and subgraph-child context snapshots as an unprovable class; kept the run path off the check and documented the limitation.
- Noted but **not** touched (pre-existing, unrelated): the tool `outputs` node attr written by the adapter is read by nothing.

## Left out

Typed variable **declarations** (string/number/bool/json) and the numeric `>`/`<` edge comparisons they enable are the explicitly out-of-scope second half of #505 and the prerequisite for #504 — not built here. This PR lands the reachability check only.

https://claude.ai/code/session_01RuhYJxVgQPsfuHZpAYbeed

---
<a href="https://app.blacksmith.sh/2389-research/codesmith/tracker/pr/513"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787861063&installation_model_id=21126&pr_number=513&repository=2389-research%2Ftracker&return_to=https%3A%2F%2Fgithub.com%2F2389-research%2Ftracker%2Fpull%2F513&signature=28c90ae81ed44344869f7f4000babf048ff715bd452faaee8b027bfcc2fa896b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added submit-time validation for unavailable context variables across conditions, interpolations, and declared reads.
  * Added a reusable validation helper for pre-acceptance checks.
  * Validation now identifies unknown references before execution, helping prevent misrouting and empty runtime values.

* **Bug Fixes**
  * Improved handling of graph reachability, loops, built-in variables, and opaque workflow components.

* **Documentation**
  * Documented variable-availability validation rules and behavior.


clintecker:
Rebased onto main (which now carries #509/#510 and #511) and **moved the hook from `tracker.ValidateSource` into `pipeline.ValidateAll`**.

Reason: #510 tightened the file-size ratchet baseline for `tracker.go` to its exact size, so the 7-line hook here pushed it over. Re-measuring wasn't an option — the PR base now carries the tightened value, so raising it would *grow* the baseline and trip the `baseline-shrinks` check.

`ValidateAll` turns out to be the better home regardless. It has exactly two callers — `tracker.ValidateSource` (the library submit seam) and the conformance validator — and is deliberately **not** on the run path, which goes through `Validate`/`validateGraph`. So the check still reaches every submit path, still cannot break a run on a key it can't prove, and now travels with the validator instead of being bolted on at the library boundary.

Also ran an independent false-positive sweep over the 13 parseable `.dip` fixtures **outside** `examples/` (conformance goldens, `testdata/`) that the original verification didn't cover: zero errors, zero warnings.

CHANGELOG conflict with the #509 and #511 entries resolved by keeping all three.

Issue #518, opened 2026-07-29
Cost reporting: no source of truth for pricing, and no way to detect drift


sugi-chan:
## Problem

Cost figures in `run.json`, the CLI summary, and the `--max-cost` budget ceiling are not currently trustworthy, and nothing in the repo detects when they drift. Pricing lives as hand-entered constants in `llm/catalog.go` plus two multipliers in `llm/pricing.go`, with no link to a source of truth and no staleness signal.

Four separate defects were found by hand in a single session, all of which produced plausible-looking numbers:

| Defect | Effect | Was it visible? |
|---|---|---|
| Cache writes priced at 0.25x input instead of 1.25x | Understated every cached Anthropic run's cache-write cost by 5x — ~1/3 of the bill on a caching-heavy run | No |
| Gemini `thoughtsTokenCount` never decoded | Billed thinking tokens counted as zero; on a reasoning-heavy call that is most of the output | No |
| OpenAI `cached_tokens` never decoded | Cached input charged at the full rate | No |
| Cached-read multiplier assumed provider-wide | GPT-4.1 (0.25x) and GPT-4o (0.5x) families underpriced against a hardcoded 0.1x | No |

Fixed in 29bdf55 and cc757b1. The concern is the ones we haven't found.

## Why the existing tests couldn't catch these

- **Token counts stayed correct and the arithmetic stayed self-consistent.** The manifest agreed with the CLI summary because both read the same wrong constant.
- **The unit test had the wrong multiplier baked into its expectation and its comment**, so it locked the bug in rather than catching it.
- **A token-conservation check doesn't help.** We added one (every token the provider counted lands in a bucket we price) and it passes cleanly on a wrong *rate* — it only catches a dropped token *class*.

A wrong rate is only detectable by comparing against the provider's published price. Nothing does that today.

## Proposed methodology

1. **Assert against published prices, not derived constants.** Each catalog entry carries the published per-million input/output/cached prices and the URL they came from; the test computes multipliers from those rather than hardcoding them. A provider price change then fails the build with a pointer to the page to re-check. (Started in `llm/pricing_cache_test.go`.)
[...]

Issue #508, opened 2026-07-28
turn_metrics events carry empty Model/Provider and zero Usage


clintecker:
**Observed in v0.46.0**, driving a 6-agent pipeline (all `claude-sonnet-4-6`, provider `anthropic`) via an embedded pipeline run. Captured from the event stream (events.ndjson).

All 7 `turn_metrics` events had empty attribution and zero usage:
```
Model='' Usage={"input_tokens": 0, "output_tokens": 0, "total_tokens": 0}
   (x7, identical)
```

Meanwhile the corresponding `llm_finish` events carried the real data:
```json
{"Type":"llm_finish","NodeID":"EchoFreeform","Turn":1,"Provider":"anthropic",
 "Model":"claude-sonnet-4-6","FinishReason":"stop",
 "Usage":{"input_tokens":3,"output_tokens":46,"total_tokens":49,"cache_write_tokens":1561}}
```

**Why it matters:** a consumer that reads per-turn usage from `turn_metrics` (a reasonable reading of the name) silently gets zeros — it looks like a run cost nothing. We hit this building cost rollups on the event stream.

**Impact: low** — the workaround is to aggregate from `llm_finish`, which is complete. But `turn_metrics` currently looks half-wired.

**Ask:** either populate `turn_metrics.{Provider,Model,Usage}`, or document that `llm_finish` is the authoritative per-turn usage event and clarify what `turn_metrics` is for (timing only?). Documentation would fully resolve this for us.

clintecker:
Additional data point from tracker-runner usage ingestion: it is not only `turn_metrics` — the **`session_start` event also carries `Model:"" Provider:""`** in the same run (verified in events.ndjson). So both events a consumer would naturally read to attribute a session's model are empty; the correct model only appears on `llm_finish`/`llm_provider_raw`. This forced our rollup to record model="unknown" until we re-sourced from llm_finish. Same ask: populate Model/Provider on session_start + turn_metrics, or document llm_finish as the sole authoritative source.

PR #519, opened 2026-07-29
Run capture: store the executed spec and every provider request, plus a run.json manifest

sugi-chan:
Makes a tracker run reconstructable after the fact: the executed spec, every request sent to a provider, per-call/turn/tool identity, and a `run.json` manifest that indexes it all. Then fixes the pricing and usage-decoding bugs that live provider runs exposed along the way.

Verified against **live Anthropic, OpenAI, and Gemini APIs** — not just stubs. `@clintecker` flagging you since you've been deepest in this; happy to split it if the review surface is too wide.

---

## Why

We want to learn from tracker runs after they finish — which spec phrasing led to which outcome. That needs three things a run didn't previously record: **the spec that actually executed**, **what each model was actually asked**, and **a cheap index** so you can find the runs worth opening without reading gigabytes of logs.

Concretely, before this branch: the activity log stored event summaries with 80-char previews, agent-session events carried no session/turn identity (so concurrent parallel branches were indistinguishable once interleaved into one file), tool arguments were truncated, and nothing recorded which `.dip` produced the run. Dippin expands subgraphs inline at compile time, so the authored source isn't the executed graph — storing only the `.dip` would have left you reconstructing against the wrong topology.

---

## What changed

**Capture (feat commits).** Session/turn identity, node kind, attempt number, and terminal status onto disk. Tool calls attributed to the turn that issued them, with untruncated inputs. LLM events grouped by a generated `call_id`, and payloads no longer clipped. Verbatim wire request bodies from all four adapters — what actually went out after provider translation and `ProviderOptions` merging, which is what you need to reproduce a call. Executed spec (`workflow.dip` + `workflow.ir.json` + input files) stored per run.

**`run.json`.** A manifest assembled from the log at run end: goal, terminal status, per-node kind/attempts/outcome/turns/usage, run totals, tool-call counts, event counts, spec identity. `tracker run-json -w <dir> <runID>` backfills it for archived runs. Measured: 742 MB of logs across 61 archived runs collapse to 134 KB of manifests; a 39-node/54 MB run assembles in 0.57s.

**Fixes found by running it live** (the interesting half — details below).

**Not included:** a run viewer built against these logs. It lives outside this repo; happy to contribute it separately if useful.

---

## The bugs live runs exposed

These are worth reviewing carefully because most produced *plausible* numbers, which is why they survived a green test suite.

**1. `run.json` reported 3x its real token usage.** The log states the same consumption at three granularities — `llm_finish` (the call), `turn_metrics` (the turn), `decision_outcome` (the node rollup) — and the accumulator summed all three. Its `call_id` grouping only collapsed the pair it was written for; in a real agent run only `llm_finish` carries a `call_id`, so 10 of 18 usage lines fell through to a per-line key and each counted again. t6 reported 6729 input tokens against a true 2243, and 18 LLM calls against 8. Tracker's own CLI summary printed the correct figure, which is what made the manifest obviously wrong.

Fixed by taking each metric from the most direct tier present rather than summing across tiers (`pipeline/runusage.go`). Tokens and cost select independently because no tier carries both: `llm_finish` meters tokens without pricing them, `turn_metrics` prices a turn without restating cache figures.

Two things that fix must not break, both found by checking rather than assuming: node rollups fire once per node *visit*, so a retried node reports twice and both are real (keying that tier by node id discarded every visit but the last, halving 15 archived runs); and `llm_calls` can't come from the usage tier at all, whose entries are coarser than calls — counting completion events by timestamp union inflated one archived run from 86 to 161, while trusting the agent path alone saw 84 of another run's 253.

**2. Cache writes priced at 0.25x input instead of 1.25x.** Someone read Anthropic's "+25% premium" as "0.25x rate". Every cached Anthropic run understated its cache-write cost by 5x — a third of the bill on t6 ($0.0608 reported against ~$0.078 actual).

Checking the pricing pages showed cached *reads* aren't a provider-wide convention either: 0.1x on Anthropic/Gemini/GPT-5, **0.25x on GPT-4.1**, **0.5x on GPT-4o**. So the multipliers moved onto `ModelInfo` with per-model overrides. Verifying mattered — a secondary source put Gemini's cached reads at 25%, but Google publishes 10%, so the existing read multiplier was already right there.

**3. Gemini dropped billed thinking tokens.** `geminiUsageMeta` decoded only prompt/candidates/total. Google's docs confirm `totalTokenCount = prompt + thoughts + candidates` and that `candidatesTokenCount` **excludes** thoughts — which Gemini bills at the output rate. Provable from t8's own arithmetic: per-call tokens summed to 2453 against a reported total of 2527, a 74-token gap. Only ~3% there, but a direct curl on the same model returned 1 candidate token against 44 thinking tokens, so on a reasoning-heavy call **98% of billed output can vanish**. That silently defeats `--max-cost`.

**4. OpenAI dropped cached input tokens**, charging cached tokens at the full rate. Also decodes `cache_write_tokens`, which the API reports and nothing read.

The nesting rules here are counterintuitive and worth a careful look, because getting one backwards swaps a missing charge for a double charge — Anthropic's `input_tokens` **excludes** the cache buckets, while OpenAI's and Gemini's prompt counts **include** them. Documented as invariants on `llm.Usage`; adapters normalize to them. I verified OpenAI's nesting empirically rather than trusting docs: two identical calls, `input_tokens: 2514` with `cached_tokens: 1792` and `total_tokens` unchanged at 2519.

**5. Three capture gaps only live runs found.** `call_id`/`request_raw` never reached the log (the stub e2e test *declared* both fields and asserted neither, so it looked like coverage); `goal` missing on single-node runs, which write no checkpoint; parallel branch nodes carried no `node_kind` because they bypass `Engine.executeNode`.

**6. `text_delta` carried no `turn_no`.** It holds the session's final assistant message — often the node's actual product, since every earlier turn is recoverable from the next request's echoed history but the last one is echoed nowhere. It was unattributable to a turn while the `turn_end` emit on the adjacent line carried one.

---

## Verification

**Live, all three providers.** Same probe task against `claude-haiku-4-5`, `gpt-5.4-nano`, and `gemini-3-flash-preview`; all pass every field. Each reconstructs its full exchange from the stored bodies despite three different request shapes (Anthropic `system`+`messages[].content[]` blocks, OpenAI `instructions`+flat `input[]`, Gemini `systemInstruction`+`contents[].parts[]`) — system prompt, the literal command the model chose, tool result, final message, and all 8 tool schemas.

**Deterministic tests, no API key** (`tracker_capture_totals_test.go`): `run.json`'s log-derived totals cross-checked against `result.Trace`, which the engine accumulates independently while running. That comparison reproduces the 3x bug with no provider. Two fixtures — a bare `agent.Completer` (proves tokens and per-node attribution, and pins that `llm_calls` is unavailable there since tracing lives in `llm.Client`) and a streaming adapter behind a real `llm.Client`, which is the stretch the earlier stub only appeared to cover. The adapter counts its own invocations and names the exact bytes it sent: two oracles the manifest can't fake.

**Pricing** (`llm/pricing_cache_test.go`): multipliers asserted against published prices with sources, since this bug class is invisible to everything else — token counts stay right and the arithmetic stays self-consistent, so the manifest agrees with the CLI and a token-conservation check passes. The pre-existing unit test had 0.25x baked into its expectation *and* its comment, locking the bug in rather than catching it; corrected to $1.95.

**Backfill:** all 61 archived runs under `~/Downloads` re-assemble with token totals byte-identical to before the usage fix, confirming the tier overlap was introduced by this branch (which added per-call usage) rather than latent in old logs. `terminal_status` comes back empty on all of them, correctly — the field didn't exist when they ran.

`make ci` green throughout: coverage 85.6%, complexity baseline untouched (212 grandfathered, 0 new), 4 core pipelines grade A, jailcheck clean.

---

## Things I'd like a second opinion on

1. **`Usage.TotalTokens` means different things per provider.** Anthropic computes `input + output` (cache excluded); OpenAI and Gemini pass the provider's own total (cache included). It isn't comparable across providers today, so anything aggregating it — budgets, dashboards — mixes units. Normalizing to "all billed tokens" changes `--max-cost` behavior for Anthropic, so I left it alone and filed #518.

2. **Cache TTL isn't modeled.** Anthropic bills a 1-hour-TTL write at 2x versus 1.25x at 5 minutes, and `llm.Usage` carries no TTL, so 1-hour runs price ~38% low. Documented at the constant; the fix is threading TTL from the adapter.

3. **Capture is CLI-only.** `pipeline.NewJSONLEventHandler`, the trace observer, and `finalizeRunCapture` are all wired in `cmd/tracker`. An embedder calling `tracker.Run` gets nothing unless it wires all three itself. I exported `JSONLEventHandler.LLMTraceObserver()` so the non-obvious rule (skip `SessionOwned` or every session call logs twice) ships with the writer, but if harvesting runs from embedders matters, a single `Config` option to enable capture is the change I'd make next.

4. **Log volume.** `request_raw` grows monotonically per turn (5137 → 17176 bytes across t6's six turns). I measured 43–70x zstd compression on this content, which is what makes storing it viable, but nothing compresses or caps it today.

5. **Permissions.** The live log is deliberately hardened — 0600 in a 0700 dir, `O_NOFOLLOW`, explicit chmod. The run-dir mirror and `run.json` are **0644 in a 0755 dir**, and now hold verbatim prompts and full tool results where they previously held truncated previews. No credentials leak (the key is an HTTP header, never in the body — I grepped to confirm), so this is content exposure, not key exposure. Related: `wip_preserve.go` runs `git add -A` on the failure path, which stages `.tracker/` into git objects unless it's gitignored, and `tracker doctor` only warns.

## Also fixed in passing

`tracker doctor` probed retired `gemini-2.0-flash` and reported a working Gemini key as a transport failure, advising key rotation. And `gemini-2.5-flash-lite` is in the catalog but 404s as "no longer available to new users" — annotated rather than removed, since archived runs that used it still need a price when their manifests are rebuilt.

## Not covered

Per-turn cost fidelity on `--backend claude-code` / `acp`. Those bypass the adapter layer entirely and estimate usage from rune counts (`ACPUsageMarker`), so there's no wire body or `call_id`, and usage is estimated rather than metered. `RunTotals.Estimated` exists to flag this; whether it propagates into `run.json` is unverified. Ollama is also untested — it should route through `openaicompat` with a base-URL override, and its cost would read $0, correct for local but indistinguishable from unpriced (#518).

---
<a href="https://app.blacksmith.sh/2389-research/codesmith/tracker/pr/519"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787955543&installation_model_id=21126&pr_number=519&repository=2389-research%2Ftracker&return_to=https%3A%2F%2Fgithub.com%2F2389-research%2Ftracker%2Fpull%2F519&signature=1eb7e24efd2dcded5d244e6ae8d559b707f6ba557f093e180bdeaa384dcdcdb6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added the `run-json` command to generate run summaries with status, node details, usage, costs, and human interventions.
  * Added persisted run specifications, expanded pipeline metadata, and richer activity logs and traces.
* **Bug Fixes**
  * Corrected cached-token pricing and improved model-specific pricing accuracy.
  * Updated the default Gemini model used for authentication checks.
* **Security/Improvements**
  * Redacted sensitive specification parameters and tightened permissions for saved artifacts.
  * Preserved complete request and tool payloads when tracing is enabled.

### DIFF
--- a/llm/google/usage.go
+++ b/llm/google/usage.go
@@ -1,0 +1,42 @@
+// ABOUTME: Translates Gemini usageMetadata into the unified llm.Usage shape.
+// ABOUTME: Thinking sits outside the candidate count; cached content sits inside the prompt count.
+package google
+
+import "github.com/2389-research/tracker/llm"
+
+type geminiUsageMeta struct {
+	PromptTokenCount     int `json:"promptTokenCount"`
+	CandidatesTokenCount int `json:"candidatesTokenCount"`
+	TotalTokenCount      int `json:"totalTokenCount"`
+	// Billed at the output rate but NOT inside CandidatesTokenCount —
+	// totalTokenCount = prompt + candidates + thoughts. Dropping it understated
+	// the bill by however much the model thought.
+	ThoughtsTokenCount int `json:"thoughtsTokenCount"`
+	// The cached slice of the prompt, already counted inside PromptTokenCount.
+	CachedContentTokenCount int `json:"cachedContentTokenCount"`
+}
+
+// extractUsage converts Gemini usage metadata to the unified Usage struct,
+// normalizing onto the llm.Usage invariants: thinking folds into OutputTokens
+// (Gemini bills it there but omits it from candidatesTokenCount), and the cached
+// prompt slice lifts out of PromptTokenCount into its own bucket so pricing can
+// discount it without counting those tokens twice.
+func extractUsage(meta *geminiUsageMeta) llm.Usage {
+	if meta == nil {
+		return llm.Usage{}
+	}
+	u := llm.Usage{
+		InputTokens:  meta.PromptTokenCount - meta.CachedContentTokenCount,
+		OutputTokens: meta.CandidatesTokenCount + meta.ThoughtsTokenCount,
+		TotalTokens:  meta.TotalTokenCount,
+	}
+	if meta.ThoughtsTokenCount > 0 {
+		v := meta.ThoughtsTokenCount
+		u.ReasoningTokens = &v
+	}
+	if meta.CachedContentTokenCount > 0 {
+		v := meta.CachedContentTokenCount
+		u.CacheReadTokens = &v
+	}
+	return u
+}
--- a/llm/openai/usage.go
+++ b/llm/openai/usage.go
@@ -1,0 +1,55 @@
+// ABOUTME: Translates OpenAI Responses API usage into the unified llm.Usage shape.
+// ABOUTME: Both cache classes and reasoning arrive nested inside the top-level counts.
+package openai
+
+import "github.com/2389-research/tracker/llm"
+
+type openaiUsage struct {
+	InputTokens  int              `json:"input_tokens"`
+	OutputTokens int              `json:"output_tokens"`
+	TotalTokens  int              `json:"total_tokens"`
+	OutputDetail *openaiOutDetail `json:"output_tokens_details,omitempty"`
+	InputDetail  *openaiInDetail  `json:"input_tokens_details,omitempty"`
+}
+
+type openaiOutDetail struct {
+	// Already counted inside OutputTokens.
+	ReasoningTokens int `json:"reasoning_tokens,omitempty"`
+}
+
+// openaiInDetail breaks down InputTokens. Both fields are nested *inside* it and
+// billed at their own rate, so both are lifted out — see translateUsage.
+type openaiInDetail struct {
+	CachedTokens     int `json:"cached_tokens,omitempty"`
+	CacheWriteTokens int `json:"cache_write_tokens,omitempty"`
+}
+
+// translateUsage converts OpenAI usage to unified format.
+//
+// OutputTokens already includes reasoning — the llm.Usage invariant — so it
+// passes through untouched. The cache buckets arrive nested inside InputTokens
+// and are lifted out: pricing sums the buckets, so leaving them in charges the
+// full rate for discounted tokens, while moving them without subtracting counts
+// the prompt twice.
+func translateUsage(u openaiUsage) llm.Usage {
+	usage := llm.Usage{
+		InputTokens:  u.InputTokens,
+		OutputTokens: u.OutputTokens,
+		TotalTokens:  u.TotalTokens,
+	}
+	if u.OutputDetail != nil && u.OutputDetail.ReasoningTokens > 0 {
+		v := u.OutputDetail.ReasoningTokens
+		usage.ReasoningTokens = &v
+	}
+	if u.InputDetail != nil && u.InputDetail.CachedTokens > 0 {
+		v := u.InputDetail.CachedTokens
+		usage.InputTokens -= v
+		usage.CacheReadTokens = &v
+	}
+	if u.InputDetail != nil && u.InputDetail.CacheWriteTokens > 0 {
+		v := u.InputDetail.CacheWriteTokens
+		usage.InputTokens -= v
+		usage.CacheWriteTokens = &v
+	}
+	return usage
+}
--- a/llm/pricing.go
+++ b/llm/pricing.go
@@ -68,7 +68,7 @@ func cacheCost(info *ModelInfo, usage Usage) float64 {
 // is actually something to price — a zero-usage call (e.g. a probe) shouldn't
 // produce a log line.
 func warnUnknownModel(model string, usage Usage) {
-	if usage.TotalTokens == 0 && usage.InputTokens == 0 && usage.OutputTokens == 0 {
+	if !usage.anyTokens() {
 		return
 	}
 	if _, already := unknownModelWarned.LoadOrStore(model, struct{}{}); already {
--- a/llm/pricing.go
+++ b/llm/pricing.go
@@ -13,35 +13,66 @@ import (
 // common for subscription-auth backends (claude-code, ACP bridges).
 var unknownModelWarned sync.Map
 
+// Cache multipliers used when a catalog entry states none. Both are the
+// published Anthropic rates, which the GPT-5 family also matches on reads.
+//
+// The write default is a *premium*, not a discount: a 5-minute cache write is
+// 1.25x base input. 1-hour writes are 2x, which this cannot express because
+// llm.Usage carries no TTL — a 1-hour-TTL run prices ~38% low, and threading
+// TTL through the adapters is the fix if that matters. The previous value here
+// was 0.25, reading "+25% premium" as "0.25x rate", which understated the
+// cache-write cost of every cached Anthropic run by 5x.
+const (
+	defaultCacheReadMultiplier  = 0.1
+	defaultCacheWriteMultiplier = 1.25
+)
+
 // EstimateCost returns the estimated dollar cost for the given model and token
 // usage. Looks up pricing from the model catalog (supports ID and aliases).
 // Returns 0 for unknown models, logging a single warning per unknown name so
 // operators notice that their --max-cost ceiling will not apply to usage
-// priced at an unknown rate. Cache read tokens are priced at 10% of input
-// rate; cache write tokens at 25% of input rate (Anthropic pricing
-// convention, applied uniformly since other providers use similar discounts).
+// priced at an unknown rate.
+//
+// Cached prompt tokens are priced from the model's own multipliers, since the
+// discount varies by model rather than by provider — see ModelInfo. Assumes
+// InputTokens excludes the cache buckets, the llm.Usage invariant every adapter
+// normalizes to.
 func EstimateCost(model string, usage Usage) float64 {
 	info := GetModelInfo(model)
 	if info == nil {
-		// Only warn when there's actually something to price — a zero-usage
-		// call (e.g. a probe) shouldn't produce a log line.
-		if usage.TotalTokens > 0 || usage.InputTokens > 0 || usage.OutputTokens > 0 {
-			if _, already := unknownModelWarned.LoadOrStore(model, struct{}{}); !already {
-				log.Printf("[llm] EstimateCost: unknown model %q (no catalog entry); returning $0 — budget --max-cost ceiling will not apply to usage priced under this model", model)
-			}
-		}
+		warnUnknownModel(model, usage)
 		return 0
 	}
 	input := float64(usage.InputTokens) / 1_000_000 * info.InputCostPerM
 	output := float64(usage.OutputTokens) / 1_000_000 * info.OutputCostPerM
+	return input + output + cacheCost(info, usage)
+}
 
-	var cacheRead, cacheWrite float64
-	if usage.CacheReadTokens != nil {
-		cacheRead = float64(*usage.CacheReadTokens) / 1_000_000 * info.InputCostPerM * 0.1
+// cacheCost prices the cache buckets, falling back to the package defaults for
+// a model that states no multiplier of its own.
+func cacheCost(info *ModelInfo, usage Usage) float64 {
+	perM := func(tokens *int, mult, fallback float64) float64 {
+		if tokens == nil {
+			return 0
+		}
+		if mult == 0 {
+			mult = fallback
+		}
+		return float64(*tokens) / 1_000_000 * info.InputCostPerM * mult
 	}
-	if usage.CacheWriteTokens != nil {
-		cacheWrite = float64(*usage.CacheWriteTokens) / 1_000_000 * info.InputCostPerM * 0.25
-	}
+	return perM(usage.CacheReadTokens, info.CacheReadMultiplier, defaultCacheReadMultiplier) +
+		perM(usage.CacheWriteTokens, info.CacheWriteMultiplier, defaultCacheWriteMultiplier)
+}
 
-	return input + output + cacheRead + cacheWrite
+// warnUnknownModel logs once per unrecognized model name. Only warns when there
+// is actually something to price — a zero-usage call (e.g. a probe) shouldn't
+// produce a log line.
+func warnUnknownModel(model string, usage Usage) {
+	if usage.TotalTokens == 0 && usage.InputTokens == 0 && usage.OutputTokens == 0 {
+		return
+	}
+	if _, already := unknownModelWarned.LoadOrStore(model, struct{}{}); already {
+		return
+	}
+	log.Printf("[llm] EstimateCost: unknown model %q (no catalog entry); returning $0 — budget --max-cost ceiling will not apply to usage priced under this model", model)
 }
--- a/pipeline/runusage.go
+++ b/pipeline/runusage.go
@@ -1,0 +1,218 @@
+package pipeline
+
+import (
+	"strconv"
+)
+
+// usageTier ranks how directly a log line reports LLM consumption.
+//
+// The same call is reported at up to three granularities: the call itself
+// (llm_finish), the turn that issued it (turn_metrics), and the node that ran
+// the turn (decision_outcome). Summing across tiers multiplies the truth — a
+// six-turn node whose every call is restated at all three levels reports three
+// times its real token count. So totals come from the most direct tier the log
+// actually contains, and coarser restatements of the same calls are ignored.
+//
+// The tiers are ordered coarsest-first so that a higher value is more direct.
+type usageTier int
+
+const (
+	tierNodeRollup usageTier = iota // decision_outcome: one line per node
+	tierTurn                        // turn_metrics: one line per agent turn
+	tierCall                        // llm_finish: one line per LLM call
+)
+
+// usageTiers lists the tiers most-direct first, which is the order totals
+// prefer them in.
+var usageTiers = []usageTier{tierCall, tierTurn, tierNodeRollup}
+
+// usageRecord is one tier's report of consumption, tagged with the node it
+// belongs to so per-node attribution falls out of the same pass.
+type usageRecord struct {
+	nodeID string
+	totals RunTotals
+}
+
+// usageIdentity classifies a usage-bearing line and names the unit it
+// describes, so repeated reports of one unit collapse instead of summing.
+//
+// Only the call tier can repeat within itself, because both log paths report
+// the same call; call_id collapses that pair. The turn and node tiers emit once
+// per turn and once per node visit, so each of their lines is its own unit — a
+// retried or revisited node emits a separate rollup per visit, and keying those
+// by node id alone would silently discard every visit but the last.
+func usageIdentity(e jsonlLogEntry) (usageTier, string) {
+	perLine := e.Timestamp + "|" + e.Type + "|" + e.NodeID
+	switch e.Type {
+	case "decision_outcome":
+		return tierNodeRollup, perLine
+	case "turn_metrics":
+		return tierTurn, perLine + "|" + e.SessionID + "|" + strconv.Itoa(e.TurnNo)
+	default:
+		if e.CallID != "" {
+			return tierCall, e.CallID
+		}
+		return tierCall, perLine
+	}
+}
+
+// hasTokens and hasCost say which metric a record carries. Tokens and cost are
+// selected independently because no single tier carries both: llm_finish meters
+// tokens without pricing them, and turn_metrics prices a turn without
+// re-reporting per-call cache figures.
+func hasTokens(r RunTotals) bool { return r.InputTokens != 0 || r.OutputTokens != 0 }
+func hasCost(r RunTotals) bool   { return r.CostUSD != 0 }
+
+// isCompletion reports whether a line marks one LLM call finishing.
+//
+// Both log paths record the same call: the client writer emits "finish" for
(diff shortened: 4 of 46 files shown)


pipeline/runusage.go
```diff
@@ -0,0 +1,218 @@
...
+
+// usageIdentity classifies a usage-bearing line and names the unit it
+// describes, so repeated reports of one unit collapse instead of summing.
+//
+// Only the call tier can repeat within itself, because both log paths report
+// the same call; call_id collapses that pair. The turn and node tiers emit once
+// per turn and once per node visit, so each of their lines is its own unit — a
+// retried or revisited node emits a separate rollup per visit, and keying those
+// by node id alone would silently discard every visit but the last.
+func usageIdentity(e jsonlLogEntry) (usageTier, string) {
```
sugi-chan, line 44: **The subtle part of this whole PR.** Only the *call* tier can repeat within itself (both log paths report the same call, collapsed by `call_id`). The turn and node tiers emit once per unit, so each of their lines is its own unit.

I got this wrong first: keying node rollups by `node_id` looked obviously right and silently halved 15 archived runs, because a retried or revisited node emits a separate rollup per visit and I was keeping only the last. Caught it by diffing 61 archived manifests before/after, not by reasoning.

pipeline/runusage.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// llmCalls returns the number of distinct LLM calls.
+//
+// Each path independently tries to record every call, so the fuller path is
+// the better count and a union across paths would double-count. Where call_ids
+// are present they identify calls exactly and agree with the per-path counts;
+// where they are absent the per-path maximum carries it, which is a lower bound
+// if one path dropped events.
+func (l *usageLedger) llmCalls() int {
```
sugi-chan, line 110: Why this can't just be `len(calls)`: the tier carrying usage is often coarser than a call, so its entry count isn't a call count. Pre-`call_id` logs have no identity to dedupe on either.

Timestamps look like the obvious fallback and aren't — the two paths record the same call microseconds apart, so pairing by timestamp inflated one archived run from 86 calls to 161. Taking the fuller path gives 86 (hand-verified against 86 `request_start` events) and 253 for another run where the agent path only saw 84.

llm/pricing.go
```diff
@@ -13,35 +13,66 @@ import (
...
+//
+// The write default is a *premium*, not a discount: a 5-minute cache write is
+// 1.25x base input. 1-hour writes are 2x, which this cannot express because
+// llm.Usage carries no TTL — a 1-hour-TTL run prices ~38% low, and threading
+// TTL through the adapters is the fix if that matters. The previous value here
+// was 0.25, reading "+25% premium" as "0.25x rate", which understated the
+// cache-write cost of every cached Anthropic run by 5x.
+const (
+	defaultCacheReadMultiplier  = 0.1
+	defaultCacheWriteMultiplier = 1.25
```
sugi-chan, line 27: This was `0.25`. Anthropic's docs give the multipliers as *5-minute writes 1.25x base input, 1-hour writes 2x, reads 0.1x* — so 0.25 appears to be "+25% premium" misread as "0.25x rate", understating cache-write cost 5x.

Known limitation: `llm.Usage` has no TTL field, so this constant can only be right for one case. 1.25x is correct for the API default; a 1-hour-TTL run prices ~38% low. Threading TTL from the adapter is the fix if that matters to anyone.

llm/openai/usage.go
```diff
@@ -0,0 +1,55 @@
...
+		OutputTokens: u.OutputTokens,
+		TotalTokens:  u.TotalTokens,
+	}
+	if u.OutputDetail != nil && u.OutputDetail.ReasoningTokens > 0 {
+		v := u.OutputDetail.ReasoningTokens
+		usage.ReasoningTokens = &v
+	}
+	if u.InputDetail != nil && u.InputDetail.CachedTokens > 0 {
+		v := u.InputDetail.CachedTokens
+		usage.InputTokens -= v
```
sugi-chan, line 46: The subtraction is load-bearing and **not symmetric with Anthropic** — this is the easiest thing to get backwards here.

OpenAI reports cached tokens *inside* `input_tokens`; Anthropic reports them *outside*. `EstimateCost` sums the buckets, so copying Anthropic's mapping (assign without subtracting) would count the prompt twice.

Verified empirically rather than from docs: two identical calls returned `input_tokens: 2514` with `cached_tokens: 1792` and `total_tokens` unchanged at 2519 — cached is a subset, and the total doesn't double-count it.

llm/google/usage.go
```diff
@@ -0,0 +1,42 @@
...
+// (Gemini bills it there but omits it from candidatesTokenCount), and the cached
+// prompt slice lifts out of PromptTokenCount into its own bucket so pricing can
+// discount it without counting those tokens twice.
+func extractUsage(meta *geminiUsageMeta) llm.Usage {
+	if meta == nil {
+		return llm.Usage{}
+	}
+	u := llm.Usage{
+		InputTokens:  meta.PromptTokenCount - meta.CachedContentTokenCount,
+		OutputTokens: meta.CandidatesTokenCount + meta.ThoughtsTokenCount,
```
sugi-chan, line 30: Folding thoughts into `OutputTokens` rather than only recording them separately, because `EstimateCost` doesn't price `ReasoningTokens` — leaving them out means Gemini bills them and we charge nothing.

Google's API reference confirms both halves: `candidatesTokenCount` excludes thoughts, and `totalTokenCount = prompt + thoughts + candidates`. Also confirms `promptTokenCount` *includes* the cached slice ("still the total effective prompt size"), hence the subtraction below.

One class I don't decode: `toolUsePromptTokenCount`. It was 0 on our runs so the fidelity check passes, but it could be non-zero with Google's built-in tools.

sugi-chan (commented):
Inline notes on the spots most likely to be misread — mostly places where the obvious reading is wrong, or where I got it wrong first and what caught it.

sugi-chan:
@clintecker — tagging you here since you've been deepest in tracker lately. If your bots want a review order, this is roughly descending risk:

1. `pipeline/runusage.go` — the tier-selection logic. Most likely place for a subtle wrong answer, and the one I already got wrong twice (see inline notes). The invariant to hold it to: node usage shares must sum to the run total, asserted in `tracker_capture_totals_test.go`.
2. `llm/{openai,google}/usage.go` + `llm/types.go` — the nesting rules. Anthropic's `input_tokens` excludes cache buckets; OpenAI's and Gemini's prompt counts include them. Wiring one by analogy with the wrong provider trades a missing charge for a double charge, so I documented them as invariants on `llm.Usage` and normalized in each adapter.
3. `llm/pricing.go` + `llm/catalog.go` — multipliers are now per-model because cached reads are 0.1x / 0.25x / 0.5x depending on family. Asserted against published prices in `llm/pricing_cache_test.go`; if a provider moves a price that test should fail and the fix is re-checking the page, not adjusting the expectation.
4. Everything else is the capture plumbing, which is mechanical field-threading and well covered.

Two specific things I'd value a second opinion on, both flagged in the description and neither of which I changed unilaterally: the `Usage.TotalTokens` cross-provider inconsistency (#518), and that all capture wiring lives in `cmd/tracker` so `tracker.Run` embedders get nothing.

Worth knowing about the test strategy, because it shaped what I trust: a stub e2e test here *declared* `call_id` and `request_raw` as fields and asserted neither, so it read as coverage while both were absent from every live run. Most of the real bugs came out of live provider runs and cross-checking two independent sources, not from unit tests — so if the bots find the tests over-fitted to the implementation, that's a fair hit and I'd rather hear it.

sugi-chan:
## ⚠️ Needs a rebase — and there's a real collision worth deciding before I resolve it

`main` moved ~20 commits since I branched, and one of them lands in exactly the same place: **28b6ce6 "fix(activity): write per-turn agent usage to activity.jsonl (audit finding 1)"**. I tried the merge, found the collision below, and backed it out rather than resolve it unilaterally — @clintecker you wrote the other side, so you should have the call.

**16 conflict hunks** across `pipeline/events_jsonl.go` (3), `pipeline/events_jsonl_test.go` (8), `cmd/tracker/run.go` (2), `agent/session.go` (1), `agent/session_run.go` (1), `pipeline/handlers/parallel.go` (1). Most are mechanical. Two are not:

### 1. We added the same data under different field names

Both branches write per-turn cache/cost onto agent `turn_metrics` / `llm_finish` lines:

| | main (28b6ce6) | this branch |
|---|---|---|
| cache read | `token_cache_read` | `cache_read_tokens` |
| cache write | `token_cache_write` | `cache_write_tokens` |
| per-turn cost | `turn_cost_usd` | `estimated_cost` |

**My read: main's names should win** — they shipped first and mirror the NDJSON `StreamEvent` fields from #508, which is the better consistency argument. That means dropping my three and rewiring `pipeline/runusage.go` to read main's.

That rewiring is the part I don't want to do carelessly. `runusage.go` selects which granularity tier to trust; if it reads a field name that no longer exists it silently sees zero and under-reports, which is *exactly* the failure mode this PR exists to fix. I got the tier logic wrong twice already and both times only a 61-run before/after diff caught it. So I'd rather land that on top of a settled naming decision, with the archived-run diff and the live three-provider ladder rerun as the check, than bury it in a merge commit.

### 2. `jsonlLogEntry` moved

I relocated the struct to `pipeline/events_jsonl_entry.go` because `events_jsonl.go` was sitting exactly at its 578-line complexity-ratchet ceiling and my fields wouldn't fit. Main has since added gate-lifecycle fields (`gate_id`, `gate_mode`, … #509) to it in place. Resolution is to keep it in the new file and port main's additions across — mechanical, but it touches every hunk in `events_jsonl_test.go`.

### Question for you

Does 28b6ce6 make part of this PR redundant? If per-turn usage now persists on main, some of my capture plumbing may be duplicative, and it's plausible the tier-selection logic in `runusage.go` gets simpler against post-28b6ce6 event shapes. I'd rather cut what's redundant than merge both and reconcile after.

**The PR as pushed is complete and green** (`make ci` passing at 29bdf55, verified live against Anthropic/OpenAI/Gemini) — it just needs this rebase. Everything in the description and inline notes still stands; the bugs and their evidence are unaffected. Happy to do the resolution as soon as the naming question is settled, or to split the pricing/usage-decoding fixes (`llm/`) out as a separate smaller PR since those are independent of the activity-log collision and probably want to land regardless.

sugi-chan:
Worked all 7 of CodeRabbit's actionable findings plus 2 nitpicks — pushed as 527c4af. **I verified each against the code before touching it, and all 7 were real.** Notes where my conclusion differs from the report:

| Finding | Verdict | Fix |
|---|---|---|
| `runusage.go` last-writer-wins drops node attribution | **Real, best catch of the review** | `keepKnownNode` carries a known node id onto a replacement that lacks one |
| `--param` values persisted unredacted | **Real, security** | Redact values under secret-looking keys; keys kept |
| `recordBundleIdentity` never called | **Real, impact overstated** | Wired on both run paths — see below |
| Per-node engine usage assigned, not accumulated (my test) | **Real** | Accumulate |
| Warning suppressed for cache-only usage | **Real** | `Usage.anyTokens()` counts cache buckets |
| `RequestRaw` doc names the wrong event | **Real** | Corrected to `EventRequestSent` |
| `digest(x) != digest(x)` trips SA4000 | **Real** | Bind the first call |
| Nitpick: spec artifacts `0o644` | Taken | → `0o600`; compounds the param exposure above |
| Nitpick: stale fixture name in comment | Taken | Fixed |

**On `recordBundleIdentity`:** the report says `run.json` "will never carry the `.dipx` identity." Not quite — `applyCheckpointToManifest` sets it from the checkpoint (`runmanifest.go:362`), and `m.Spec.BundleIdentity` falls back to that. So `run.json` was fine; only `SpecManifest.BundleIdentity` was empty at write time. The function was still genuinely dead code, so it's now called on both the non-TUI and TUI paths.

**On the node-attribution one** — this is the finding I'm most glad it caught, and it lands squarely on the file I flagged as highest-risk in my own review notes. It's latent on the CLI path (`LLMTraceObserver` skips `SessionOwned` events, so the node-less duplicate doesn't currently get written) but reachable by any embedder wiring `WriteLLMEvent` directly. Exactly the failure mode this PR exists to prevent: run totals stay correct while the per-node shares quietly stop summing to the whole. Regression test added (`TestNodeAttributionSurvivesANodelessDuplicate`) that fails without the fix.

**On the param redaction** — worth being explicit that key-name matching is a floor, not a guarantee: a credential passed under an innocuous name still persists. Said so in the comment. The durable fix is keeping secrets out of `--param` entirely, and this pairs with the `0o644`→`0o600` change since both touch the same exposure.

Regression tests added for the two behavioral fixes. `make ci` green — complexity ratchet still at **0 new violations** (needed two small helper extractions, `Usage.anyTokens` and `keepKnownNode`, to stay there).

⚠️ **Still blocked on the rebase question in my earlier comment** — the collision with 28b6ce6 over per-turn usage field naming (`token_cache_read` vs `cache_read_tokens`). @clintecker that's the one thing I don't want to resolve unilaterally, since rewiring `runusage.go` to read renamed fields is exactly where a silent zero would hide. Everything above is independent of it.

sugi-chan:
Round 2 finding fixed in bef8f97 — and it was a genuine hole in my round 1 fix, not a nitpick.

**`os.WriteFile` applies its mode only when it creates the file.** Dropping the literal to `0o600` did nothing for a run directory that already held a `0644` artifact — a resumed run, a reused dir. The spec carries the run's `--param` values and whatever reference material it was given, so those stayed readable by other local users while the code read as though they didn't.

What makes it a fair hit: **this repo already knew the lesson.** `events_jsonl.go` force-tightens with an explicit `os.Chmod` after `OpenFile`, and the comment there spells out exactly this reasoning ("the 0o600 in OpenFile only applies at creation, so a pre-existing file from a prior run … could retain wider permissions"). I changed a mode literal and didn't follow the established pattern two files over.

Now routed through a shared `writeTightened` helper so all three artifact paths get it, with `specFileMode` named once. The chmod is best-effort by design — the write already landed, and failing a run over a mode we couldn't narrow would trade a real artifact for a defense-in-depth measure.

Regression test pre-creates `workflow.dip`, `workflow.ir.json`, and an `inputs/*` entry at `0644` and asserts all three come back `0600`. **I verified it fails without the chmod** rather than assuming — it reports all three files, so it's testing the mechanism and not passing incidentally.

`make ci` green, complexity ratchet still 0 new.

---

Status for anyone picking this up: **all 8 CodeRabbit findings across both rounds are resolved.** The only thing outstanding is the rebase question from my earlier comment — the collision with 28b6ce6 over per-turn usage field naming (`token_cache_read` vs `cache_read_tokens`). @clintecker that's still the one call I don't want to make unilaterally, since rewiring `runusage.go` to read renamed fields is precisely where a silent zero hides, and the archived-run diff is what caught me being wrong there twice. Everything else on this branch is independent and green.

clintecker:
Landed on `main` via an integration branch (merge commit 4ac8249) — thank you, @sugi-chan. This is a lot of careful work, and the live-provider verification on the four cost bugs is exactly why they were worth catching.

It didn't merge cleanly because three releases (v0.47.0/48/49) rewrote the same event/session/capture plumbing in the interim (per-turn `turn_metrics` attribution #508, per-turn agent usage → activity.jsonl, `run_id` on handler events, the gate lifecycle + buffered-handler work, the diagnostics sink, the tool guardrail). I resolved the 6 conflicts preserving both sides and verified each survives with a witness test: your `TestRunCapture_ActivityLogCarriesReconstructionFields` and our `TestParseActivityLine_AgentTurnUsageSurvives` + `TestCacheWriteIsAPremiumNotADiscount` all pass on the merged tree. Full suite, complexity ratchet, parity guards, API-surface snapshot, and golden traces are green; `dippin doctor` A/100 on the core pipelines.

Two integration notes:
- Your `WriteAgentEvent(evt agent.Event)` signature (a superset) replaced our `AgentTurnUsage` parameter form; the per-turn-usage data + field names are preserved via `applyAgentUsage`, and the redundant `cmd/tracker/agent_usage.go` glue was removed.
- `jsonlLogEntry` now carries per-turn economics under two name sets (our `token_cache_*`/`turn_cost_usd` and your `cache_*_tokens`/`estimated_cost`), both consumed. Filing a follow-up to converge them.

The broader pricing-source-of-truth methodology you raised in #518 stays open as the next workstream.

